Issue #2, opened 2020-10-13
compe_nvim_lsp/source.lua:60: attempt to index local 'result' (a nil value)


tami5:
Hi @hrsh7th,
I got to say, after going through many completion plugins yours offered the best experience. am amazed of how light and fast it feels, also NO FLICKERING which is so amazing.

During testing i got the following error whenever I do one of the following
- type ./... to complete file path
- after completing lsp source function and opening paren (..

```viml
Error executing vim.schedule lua callback: ...pack/packer/opt/nvim-compe/lua/compe_nvim_lsp/source.lua:60: attempt to index local 'result' (a nil value)
```

My configuration:

```lua
  vim.o.pumheight = 10
  vim.cmd('set completeopt=menuone,noinsert,noselect')
  vim.cmd('packadd nvim-compe')
  vim.cmd'let g:compe_enabled = v:true'
  vim.cmd'let g:compe_min_length = 1'
  require'compe_nvim_lsp'.attach()
  require'compe':register_lua_source('buffer', require'compe_buffer')
  vim.cmd[[call compe#source#vim_bridge#register('path', compe_path#source#create())]]
  bind.i({
    {'<cr>', [[compe#confirm('<CR>')]], {expr = true}},
    {'<C-e>', [[compe#close('<C-e>'))]], {expr = true}},
    {'<C-Space>', [[<C-r>=compe#complete()<CR>]], {expr = true}},
    {'<C-j>', [[pumvisible() ? "\<C-n>" : "\<C-j>"]], { expr = true}},
    {'<C-k>', [[pumvisible() ? "\<C-p>" : "\<C-k>"]], { expr = true}},
    {'<Tab>', [[pumvisible() ? "\<C-n>" : "\<Tab>"]], { expr = true}},
    {'<S-Tab>', [[pumvisible() ? "\<C-p>" : "\<S-Tab>"]], { expr = true}},
  })

```

I'm using nvim built in lsp 

Seriously amazing plugin and I believe it might be the next best thing. Looking forward to develop sources for it

tami5:
Also, after the above error message going to insert is no longer allowed: `E523: Not allowed here` error

tami5:
path completion doesn't work, typing ./ does trigger path completion

hrsh7th:
path completion problem should be fixed now.

tami5:
It is, thanks @hrsh7th

tami5:
[...]

PR #3, opened 2020-10-13
fix(nvim_lsp): attempt to index local result L60

tami5:
possible fix of #2 . After adding those lines and few test, I no longer have the aforementioned error message. Not sure if its the right solution to the issue.

### DIFF
--- a/lua/compe_nvim_lsp/source.lua
+++ b/lua/compe_nvim_lsp/source.lua
@@ -2,6 +2,8 @@ local Pattern = require'compe.pattern'
 
 local Source = {}
 
+Source.callback = false
+
 function Source.new(client)
   local self = setmetatable({}, { __index = Source })
   self.client = client
@@ -47,6 +49,18 @@ end
 
 function Source:complete(args)
   local params = vim.lsp.util.make_position_params()
+  local buf_has_clients = function()
+    for _ in pairs(vim.lsp.buf_get_clients()) do
+      return true
+    end
+    return false
+  end
+
+  if not buf_has_clients() then
+    Source.callback = true
+    return
+  end
+
   params.context = {
     triggerKind = (args.trigger_character_offset > 0 and 2 or (args.incomplete and 3 or 1))
   }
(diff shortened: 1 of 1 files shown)


lua/compe_nvim_lsp/source.lua
```diff
@@ -47,18 +49,32 @@ end
...
   local params = vim.lsp.util.make_position_params()
+  local buf_has_clients = function()
+    for _ in pairs(vim.lsp.buf_get_clients()) do
+      return true
+    end
+    return false
+  end
+
+  if not buf_has_clients() then
+    Source.callback = true
```
hrsh7th, line 60: You should call `return args.abort()` and remove `Source.callback` property.

lua/compe_nvim_lsp/source.lua
```diff
@@ -47,18 +49,32 @@ end
...
   params.context = {
     triggerKind = (args.trigger_character_offset > 0 and 2 or (args.incomplete and 3 or 1))
   }
   if args.trigger_character_offset > 0 then
     params.context.triggerCharacter = args.context.before_char
   end
 
-  self.client.request('textDocument/completion', params, function(method, err, result)
+  self.client.request('textDocument/completion', params, function(err, _, result)
+    if err or not result then Source.callback = true return end
```
hrsh7th, line 72: Same.

hrsh7th:
Thank you.
I've added few comments.

lua/compe_nvim_lsp/source.lua
```diff
@@ -47,18 +49,32 @@ end
 
 function Source:complete(args)
   local params = vim.lsp.util.make_position_params()
+  local buf_has_clients = function()
```
hrsh7th, line 52: I think should be good to change to `if vim.lsp.client_is_stopped(self.client.id) then return args.abort() end`.

tami5:
Done

hrsh7th:
LGTM. Thank you!

tami5:
No thank you @hrsh7th, i'm so amazed of how awesome it feels using it. You did a great job honestly

Issue #30, opened 2020-11-24
backspace causes flickering 


tami5:

hrsh7th:
Maybe implemented it.

tami5:
I’ll test it later and report back. Thanks @hrsh7th

elianiva:
hmm, I got an issue after this commit.
https://streamable.com/38nu2c

Notice the last few seconds, it's not me deleting `d` and typing it again, I can't delete the previous word if the popup is visible. Here's my [config](https://github.com/elianiva/dotfiles/blob/master/nvim/.config/nvim/lua/plugins/_completion.lua) for nvim-compe.

hrsh7th:
@elianiva Thank you for your report.
I've fixed it maybe.

elianiva:
Can confirm that it's gone, but I still can't delete the previous word unfortunately.

hrsh7th:
It's very weird and I can't reproduce.

Could you provide a result of `imap <BS>` ?

elianiva:
It gives me this.
![image](https://user-images.githubusercontent.com/51877647/100205862-3a6f9f80-2f38-11eb-9a45-a54de2eb8e2a.png)

hrsh7th:
Sorry. Could you test with `:imap <BS>` ?

elianiva:
![image](https://user-images.githubusercontent.com/51877647/100207054-b4ecef00-2f39-11eb-9996-8b29fd25d45c.png)
ok, it gives me this.

hrsh7th:
Oh... It is the same result as mine...
I will consider it

tami5:
@elianiva why use this ? doesn't completion on backspace works naturally without this mapping? or is there additional stuff it brings?

elianiva:
I didn't set that mapping. It came from lexima.

tami5:
ohhhh I see

kristijanhusak:
@hrsh7th I noticed from commit https://github.com/hrsh7th/nvim-compe/commit/0ccef7c6558a0f00adea4e892c6ec58495f6fc3a that "tags" source is misbehaving. It won't properly show results util you write out the whole thing, and even then gives wrong results. Commit before that one works fine.

hrsh7th:
@kristijanhusak Thank you for your report.

This behavior is intentional.
For example, the buffer source should be invoked once on one alphabetical char typed (If we gather words on each key press, It isn't efficient because buffer words would not change during one completion.

[...]

PR #35, opened 2020-11-25
Fix tags completion to always use fresh input.

kristijanhusak:
Fixes issue mentioned in [this comment](https://github.com/hrsh7th/nvim-compe/issues/30#issuecomment-733608801)


hrsh7th:
Thank you!

PR #44, opened 2020-11-30
[WIP] Improve setup implementation

hrsh7th:
### Overview

1. the source must be registered by the `source author`
    - It can be registered after setup.
    - The source will be detected on each keypress.

2. the source must be configured by the `user`.
    - So the user should install the source and add it to the configuration like the below.

```viml
" The source author must register the source automatically.
Plug 'some-source-author/compe-awesome-source'

" The user must configure source manually to enable it.
let g:compe = {}
let g:compe.source = {}
let g:compe.source.awesome_source = v:true
```

### Breaking change

1. `compe#source#vim_bridge#register` is not work now.
- Use `compe#register_source(name, source)` instead.

2. `require"compe":register_lua_source` is not work now.
- Use `require"compe":register_source(name, source)` instead.

3. The source does not be enabled automatically. To enable it needs explicit configuration.
- It will have to do by the user.

### DIFF
--- a/lua/compe/completion/init.lua
+++ b/lua/compe/completion/init.lua
@@ -1,5 +1,6 @@
 local Debug = require'compe.debug'
 local Async = require'compe.async'
+local Config = require'compe.config'
 local Context = require'compe.completion.context'
 local Matcher = require'compe.completion.matcher'
 local VimBridge = require'compe.completion.source.vim_bridge'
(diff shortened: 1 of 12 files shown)


hrsh7th:
I think it is good change to organize some API's.

So I will refactor som codebase before merge this.

hrsh7th:
I will refactor some utilities move to `lua/compe/utils` directory.
And remove `lua/compe/init.lua`.

lua/compe/completion/init.lua
```diff
@@ -254,6 +242,25 @@ function Completion.should_ignore_display(self)
...
+    end
+  end
+
+  table.sort(sources, function(source1, source2)
+    local meta1 = source1:get_metadata()
+    local meta2 = source2:get_metadata()
+    if meta1.priority ~= meta2.priority then
+      return meta1.priority > meta2.priority
+    end
+  end)
```
hrsh7th, line 260: It should cache.

tami5: What you mean?

hrsh7th: The get_sources is called frequently.

tami5: Oh I see

PR #45, opened 2020-12-01
Refactor

hrsh7th:


tami5:
Finally :heart:

hrsh7th:
This PR has a lot of breaking change so I will create PR for external sources.

I've defined the Public API.
I plan to keep it after this PR.
So the Public API signature is super important.

Please comments for this PR if someone don't like Public API's signature,

hrsh7th:
I'm considering the API to get regexp pattern from custom source.

tami5:
> This PR has a lot of breaking change so I will create PR for external sources.
> 
> I've defined the Public API.
> I plan to keep it after this PR.
> So the Public API signature is super important.
> 
> Please comments for this PR if someone don't like Public API's signature,

Does this pr include #44 ? Also it would be cool if this can be located in develop branch where other can test it until its mature then it can be merge to master

hrsh7th:
Yes. It' included #44.

OK. I will create dev branch.

tami5:
Nice I will switch to it in few days, it good to sync it with a this pr.

PR #58, opened 2020-12-25
WIP: add border to documentation.

tami5:
Trying to add border to documentation window. Something is wrong with sizing. but it works as expected.

The background highlight of the popup can be adjusted via `vim.api.nvim_win_set_option` with win ids and setting `winhl` to a highlight, e.g. "NormalFloat:Normal". I think this need to be optional after fixing the sizing.

Also I noticed that the window get created and destroyed on each item. It might be a good idea to update the buffer content instead via setlines and using buffer ID.


hrsh7th:
It's looks like a good approach to me.

Issue #31, opened 2020-11-24
Implement border, padding and some pretty staff in documentation.


hrsh7th:
It should be implemented in https://github.com/hrsh7th/vim-vital-vs

tami5:
Thats amazing @hrsh7th , are you going to copy the code over here?

hrsh7th:
I'm using vital.vim.
It has embeddable module system.

tami5:
Just got an overview of vital.vim, seems interesting. But just to make sure nothing in  https://github.com/hrsh7th/vim-vital-vs are duplicated here, right?

hrsh7th:
Yes. vital.vim is `embeddable` library so I embedded some modules to vim-vsnip, nvim-compe, vim-lamp and vim-candle.

tami5:
Oh nice. seems to be very interesting

PR #59, opened 2020-12-26
Implement border

hrsh7th:
#58 is good approach to support border for documentation.

But it is bit difficult because the implementation relates `vim-vital-vs` too.

So I implement it with referencing #58.

TODO and other concerns.

1. It should be an option?
2. border settings.
3. we should use `strdisplaywidth` for calculate positions.

https://user-images.githubusercontent.com/629908/103150434-891e8c80-47b7-11eb-8c7b-b8fffdf4d2e4.mp4

### DIFF
--- a/autoload/compe/documentation.vim
+++ b/autoload/compe/documentation.vim
@@ -43,7 +43,7 @@ function! compe#documentation#open(document) abort
   \   'col': l:layout.col,
   \   'maxwidth': l:maxwidth,
   \   'maxheight': l:maxheight,
-  \   'filetype': '',
+  \   'filetype': 'border',
   \   'contents': l:borders,
   \   'winhl': 'NormalFloat:Normal',
   \ })
--- a/autoload/compe/documentation.vim
+++ b/autoload/compe/documentation.vim
@@ -2,36 +2,59 @@ let s:MarkupContent = vital#compe#import('VS.LSP.MarkupContent')
 let s:FloatingWindow = vital#compe#import('VS.Vim.Window.FloatingWindow')
 
 let s:window = s:FloatingWindow.new()
+let s:border = s:FloatingWindow.new()
 
 "
 " compe#documentation#show
 "
 function! compe#documentation#open(document) abort
+  call compe#documentation#close()
   if !compe#_is_selected_manually()
     return
   endif
 
   let l:document = split(s:MarkupContent.normalize(a:document), "\n", v:true)
-
-  let l:pos = s:get_screenpos(pum_getpos(), l:document)
-  if empty(l:pos)
-    return s:window.close()
+  let l:layout = s:get_screenpos(pum_getpos(), l:document)
+  if empty(l:layout)
+    return
   endif
 
+  let l:maxwidth = float2nr(&columns * 0.4)
+  let l:maxheight = float2nr(&lines * 0.4)
+
   call s:window.open({
-  \   'row': l:pos[0],
-  \   'col': l:pos[1],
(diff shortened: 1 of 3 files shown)


hrsh7th:
https://github.com/hrsh7th/nvim-compe/issues/31

tami5:
😍😍 Well done.

> It should be an option?

Definitely, like nvim_compe.border = true or { settings }
> border settings.

The color of border, transparency maybe 

> we should use strdisplaywidth for calculate positions.

What you mean? what are we using now

... other consideration

update buffer content if next content is not empty, otherwise close.

hrsh7th:
Currently, vim-vital-vs's `s:FloatingWindow` will use a cached window.
But it remained the rendering garbage... So I choose this PR's approach (always close -> open)

autoload/compe/documentation.vim
```diff
@@ -2,36 +2,59 @@ let s:MarkupContent = vital#compe#import('VS.LSP.MarkupContent')
...
-  \   'maxheight': float2nr(&lines * 0.4),
+  \   'row': l:layout.row + 1,
+  \   'col': l:layout.col + 1,
+  \   'maxwidth': l:maxwidth - 2,
+  \   'maxheight': l:maxheight - 2,
   \   'filetype': 'markdown',
   \   'contents': l:document,
+  \   'winhl': 'NormalFloat:Normal',
+  \ })
+
```
tami5, line 34: Add something like `chars = vim.g.compe.border.chars or [defaults]` to enable use customization.

hrsh7th: Good.

```
let g:compe = {}
let g:compe.border = v:false
let g:compe.border = {}
let g:compe.border.chars = {}
let g:compe.border.chars.top_left = ''
let g:compe.border.chars.top_right = ''
let g:compe.border.chars.bottom_left = ''
let g:compe.border.chars.bottom_right = ''
let g:compe.border.chars.vertical = ''
let g:compe.border.chars.horizontal = ''
```

I don't plan to add colors etc for now.

tami5: I'd say make it an array of chars [bottom_corner_l, bottom_corner_r, top_corner_r, top_corner_l, top/bottom, left/right]

so the user can do [char, char, ... ] in specific order.

hrsh7th:
I will try it

Ethaks:
[nvim_open_win()](https://neovim.io/doc/user/api.html#nvim_open_win()) now has an option `border`, allowing for `'single'`, `'double'`, and `'shadow'`, which makes the addition of borders as easy as adding `'border': 'single'` to the style, although that alone does not take care of a proper offset to account for the changed width etc. of course.
I only glanced at the MR dealing with borders however, so if the current idea already deals with all of this on its own, feel free to ignore my comment.

cloggier:
Yes, using built in functionality should be the way to go here. It allows users to configure floating windows to look exactly the same.

sam20908:
Looking forward to this getting merged! 🥳

rcarriga:
For anyone else waiting for this, I've thrown together a quick branch with the new border in nightly: https://github.com/rcarriga/nvim-compe/tree/feat/doc-border. Obviously it's lacking any real customisation but if you're happy with defaults, it works!

hrsh7th:
For now, the nvim nightly has a border option so we can provide border more easily...
@rcarriga Thank you for your information.

sam20908:
I've tested the branch from rcarriga, and it works!

folke:
Added support for borders to #426

Issue #61, opened 2021-01-06
weird backspace behavior


RishabhRD:
Hello @hrsh7th , thanks for the really cool plugin. I shifted from completion-nvim and I feel it has the abstractions that people really want from an auto-completion setup.

Just I found the backspace behavior a little weird. When I leave some suggestion for my own text and then do a backspace it inserts the (left) suggestion text automatically. Very similar to android autocorrection :sweat_smile: 
I have attached a gif that should explain my words more clearly.
![backspace](https://user-images.githubusercontent.com/26287448/103820708-cfff6280-5092-11eb-96ec-83d67076f4d0.gif)
I don't know if it is the intended feature. If it is then is there any option to disable it.

Also, sometimes backspace deletes a big amount of text (like 2 or 3 words) instead of a character.
This gif would show the behavior:
![backmore](https://user-images.githubusercontent.com/26287448/103821041-6f245a00-5093-11eb-8453-61421274e1a7.gif)

(Bottom right of gif shows my keypress in case of any ambiguity)

Thanks

hrsh7th:
Could you provide your config?

RishabhRD:
@hrsh7th I just realized that this is due to my `completeopt`. I also set the longest parameter that was causing this behavior. Sorry for this dumb issue. :sweat_smile: Thanks again for plugin

hrsh7th:
It's weird... nvim-compe should overwrite complwteopthttps://github.com/hrsh7th/nvim-compe/blob/master/lua/compe/completion.lua#L229

RishabhRD:
That's weird. I guess that the function is not being called in case of backspace
but completion is being triggered. This maybe a possible scenario. I would
also check that.

The minimal configuration needed to regenerate the behavior is:

```vim
call plug#begin('~/.vim/plugged')
Plug 'hrsh7th/nvim-compe'
call plug#end()

lua << EOF
require'compe'.setup {
  enabled = true,
  debug = false,
  -- min_length = 1,
  preselect = 'disable',

  source = {
    path = {
	priority = 9
    },
    buffer = {
	priority = 8
    },
    vsnip = {
	priority = 9
    },
    nvim_lsp = {
	priority = 10,
	sort = false
    },
  };
}
EOF
set completeopt=menuone,longest
```

hrsh7th:
Thank you for providing minimal repro.

It's reproduced. It's very weird...

hrsh7th:
[...]

PR #62, opened 2021-01-07
call vim.fn.complete when event loop is processing GUI event

RishabhRD:
Hello @hrsh7th, Opening a PR so we can discuss in terms of code. #61 shows
weird backspace behavior.

I feel `vim.fn.complete` should be called when the GUI event is taking place in
nvim-event-loop. This PR currently adds `vim.schedule` wrapper around it. At least
this solves it in my machine. What is the effect on your environment?

If not solves in your environment I can change work over code more as I am
interested in this cool project. :smile:

### DIFF
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -231,7 +231,9 @@ end
 Completion._show = function(start_offset, items)
   local completeopt = vim.o.completeopt
   vim.cmd('set completeopt=menu,menuone,noselect')
-  vim.call('complete', start_offset, items)
+  vim.schedule(function()
+      vim.call('complete', start_offset, items)
+  end)
   vim.cmd('set completeopt=' .. completeopt)
   Completion._current_offset = start_offset
   Completion._current_items = items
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you for your contribution!

I thought almost logic is safe to neovim's fast_event, in nvim-compe...

lua/compe/completion.lua
```diff
@@ -231,7 +231,9 @@ end
 Completion._show = function(start_offset, items)
```
hrsh7th, line 231: I think we should wrap all `Completion._show` codes by `vim.schedule`.

hrsh7th:
This PR improve behavior!

lua/compe/completion.lua
RishabhRD: Yeah! This would make the behavior more predictable in terms of execution.

RishabhRD:
@hrsh7th I just pushed the code to wrap the function in vim.schedule, I guess
this is what you were talking about in the code review. Or I got it wrong? :sweat_smile:

hrsh7th:
No. You are right. LGTM!

PR #63, opened 2021-01-09
call vimscript complete only when there are items in completion list

RishabhRD:
Hi @hrsh7th, in ``Completion._show`` function, vimscript ``complete`` function
is called every time even if there are no elements in ``items``.

When items are empty, we get a ``Pattern not found`` error message. This is classic
with vim's completion mechanism. However, this can be so distracting for users
many times.

This pull request adds a condition around the complete function to be called only
when there are some elements in the ``items`` table.

This removes the ``Pattern not found`` error message significantly while typing.

### DIFF
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -228,7 +228,7 @@ Completion._show = function(start_offset, items)
   vim.schedule(function()
     local completeopt = vim.o.completeopt
     vim.cmd('set completeopt=menu,menuone,noselect')
-    if not vim.tbl_isempty(items) then
+    if not (vim.call('pumvisible') == 0 and #items == 0) then
       vim.call('complete', start_offset, items)
     end
     vim.cmd('set completeopt=' .. completeopt)
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -228,7 +228,9 @@ Completion._show = function(start_offset, items)
   vim.schedule(function()
     local completeopt = vim.o.completeopt
     vim.cmd('set completeopt=menu,menuone,noselect')
-    vim.call('complete', start_offset, items)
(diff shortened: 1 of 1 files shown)


hrsh7th:
Good catch.

But compe will use empty completion explicitly.

So I left a few comments.

lua/compe/completion.lua
```diff
@@ -228,7 +228,9 @@ Completion._show = function(start_offset, items)
   vim.schedule(function()
     local completeopt = vim.o.completeopt
     vim.cmd('set completeopt=menu,menuone,noselect')
-    vim.call('complete', start_offset, items)
+    if not vim.tbl_isempty(items) then
+      vim.call('complete', start_offset, items)
+    end
```
hrsh7th, line 233: You should use `if not (vim.call('pumvisible') == 0 and #items == 0) then`.

vim needs `call complete(1, [])` to close popup menu. so we should check `pumvisible` in here.

RishabhRD: Yeah! I didn't know about it! I was observing some abrupt behavior but I was
not sure of it. Thanks! I just added a commit regarding it. The only condition we would get ``Pattern not found`` now is when we type something that didn't match from popup menu suggestions. I don't know if we can also remove it.

hrsh7th: >The only condition we would get Pattern not found now is when we type something that didn't match from popup menu suggestions.
Yes. It is expected behavior.

You can set `set shortmess+=c` to avoid it.

hrsh7th:
LGTM!

lua/compe/completion.lua
RishabhRD: Yeah! That was so lame of me :sweat_smile:. I should have read help docs. Thanks! for help.

PR #64, opened 2021-01-10
Add documentation for tags completion that contains all file paths.

kristijanhusak:
This PR adds documentation floating window for tags completion that contains all file paths where tag is defined.
This is useful when you rely a lot on tags completion (which I do in javascript projects), and i'm not always 100% sure if certain thing exist in certain file. This helps a lot.

Here are few screenshots:

Less than 10 entries:

less_then_10_entries

10 entries is maximum. Everything else is truncated:

over_10_entries

### DIFF
--- a/autoload/compe_tags/source.vim
+++ b/autoload/compe_tags/source.vim
@@ -31,15 +31,15 @@ endfunction
 " s:documentation
 "
 function! s:documentation(args) abort
-  let word = get(a:args.completed_item, 'word', '')
-  if empty(word)
-    return
+  let l:word = get(a:args.completed_item, 'word', '')
+  if empty(l:word)
+    return a:args.abort()
   endif
-  let tags = map(taglist(word), 'v:val.filename')
-  if len(tags) > 10
-    let tags = tags[0:9] + [printf('...and %d more', len(tags[10:]))]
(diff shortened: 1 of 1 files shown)


autoload/compe_tags/source.vim
```diff
@@ -26,6 +27,21 @@ function! s:datermine(context) abort
...
+  let word = get(a:args.completed_item, 'word', '')
+  if empty(word)
+    return
+  endif
+  let tags = map(taglist(word), 'v:val.filename')
+  if len(tags) > 10
+    let tags = tags[0:9] + [printf('...and %d more', len(tags[10:]))]
+  endif
+  return a:args.callback(tags)
+endfunction
```
hrsh7th, line 43: Thank you! 
Could you add `l:` predix to local variable?
Sorry. I didn't setup CI yet.

And pleaae call `l:args.abort` instead with `return`.

hrsh7th:
It looks very useful.

autoload/compe_tags/source.vim
kristijanhusak: Fixed.

hrsh7th:
LGTM. Thank you!

PR #66, opened 2021-01-10
Remove duplicate paths from tags completion documentation.

kristijanhusak:
I noticed that there are a lot of duplicate paths for tags documentation that I added today, even on the screenshot I attached:

screenshot

This is how it looks now:

after_screenshot

There are still some duplicates since the list is not sorted, but I don't want to sort it since things should be sorted as they appear in tags file, at least that seems more natural to me. So much duplicates are generally very rare and only for keywords that occur so much.


hrsh7th:
I think it is reasonable. perf is also important.

LGTM.

Issue #65, opened 2021-01-10
Any chance of adding ultisnips builtin completion-source?


RishabhRD:
Hello @hrsh7th , currently nvim-compe supports vim-vsnip for snippet completion.

Although Ultisnips is considered to be slow, it remains to be very popular due to pre-available snippet solutions like vim-snippets. This makes it configurable in almost no time with a lot of handy snippets and hence many people still use it.

So, having Ultisnips completion builtin can interest a big user-base. (Maybe I am much opinionated on it)

The only problem with Ultisnips is, its code is not much **public-accessible** and hence something like getting a raw snippet string would result in a huge increase in the codebase. However, something like available snippets at the moment can be provided very easily without increasing the codebase drastically.

If you feel like adding it, I would submit a PR that would implement the feature.

hrsh7th:
No ploblem. Looking forward to the PR.

hrsh7th:
The PR has merged.

PR #67, opened 2021-01-11
Added Ultisnips builtin completion source

RishabhRD:
Hello @hrsh7th , as discussed in #65 , This PR adds ultisnips completion source. If something is missing from the standard, I would modify it accordingly. :sweat_smile:

### DIFF
--- a/after/plugin/compe_ultisnips.vim
+++ b/after/plugin/compe_ultisnips.vim
@@ -1,0 +1,9 @@
+if exists('g:loaded_compe_ultisnips')
+  finish
+endif
+let g:loaded_compe_ultisnips = v:true
+
+if exists('g:loaded_compe')
+  lua require'compe'.register_source('ultisnips', require'compe_ultisnips')
+endif
+
--- a/after/plugin/compe_ultisnips.vim
+++ b/after/plugin/compe_ultisnips.vim
@@ -1,4 +1,4 @@
-if exists('g:loaded_compe_ultisnips')
+if match(&runtimepath, 'UltiSnips') == -1 || exists('g:loaded_compe_ultisnips')
   finish
 endif
 let g:loaded_compe_ultisnips = v:true
--- a/lua/compe_ultisnips/init.lua
+++ b/lua/compe_ultisnips/init.lua
@@ -17,6 +17,7 @@ end
 function M:complete(args)
   if vim.fn.exists("*UltiSnips#SnippetsInCurrentScope") == 0 then
     args.abort()
+    return
   end
   local receivedSnippets = vim.call('UltiSnips#SnippetsInCurrentScope')
   if vim.tbl_isempty(receivedSnippets) then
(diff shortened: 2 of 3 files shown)


after/plugin/compe_ultisnips.vim
```diff
@@ -0,0 +1,9 @@
+if exists('g:loaded_compe_ultisnips')
+  finish
+endif
+let g:loaded_compe_ultisnips = v:true
+
+if exists('g:loaded_compe')
```
hrsh7th, line 6: Could you check ultisnips installed or not?

lua/compe_ultisnips/init.lua
```diff
@@ -0,0 +1,53 @@
...
+
+function M:datermine(context)
+  return compe.helper.datermine(context)
+end
+
+function M:complete(args)
+  if vim.fn.exists("*UltiSnips#SnippetsInCurrentScope") == 0 then
+    args.abort()
+    return
+  end
```
hrsh7th, line 21: It should be moved to datermine or just removed.

lua/compe_ultisnips/init.lua
```diff
@@ -0,0 +1,53 @@
...
+  local receivedSnippets = vim.call('UltiSnips#SnippetsInCurrentScope')
```
hrsh7th, line 22: Please use snake_case instead.

lua/compe_ultisnips/init.lua
```diff
@@ -0,0 +1,53 @@
...
+  if vim.tbl_isempty(receivedSnippets) then
+    args.abort()
+    return
+  end
+  local completionList = {}
+  for key, value in pairs(receivedSnippets) do
+    local item = {
+      word =  key,
+      abbr =  key,
+      userdata = value,
```
hrsh7th, line 32: it should be `user_data` because vim's completed-item uses it.

lua/compe_ultisnips/init.lua
```diff
@@ -0,0 +1,53 @@
...
+end
+
+function M:documentation(args)
+  local completedItem = args.completed_item
+  local userdata = completedItem.userdata
+  if userdata == nil or userdata == '' then
+    args.abort()
+    return
+  end
+  args.callback(userdata)
```
hrsh7th, line 50: LGTM.

after/plugin/compe_ultisnips.vim
RishabhRD: Yeah! I would include the check using ``match`` function of vimscript

lua/compe_ultisnips/init.lua
RishabhRD: OK! I didn't know about it. Just changing it. :smiley:

lua/compe_ultisnips/init.lua
RishabhRD: Sure! :sweat_smile:

lua/compe_ultisnips/init.lua
RishabhRD: Once checked if ultisnips is installed or not while registering I don't think there would be any use for it. So, I would remove it.

RishabhRD:
@hrsh7th I have added the changes you asked for. Please review it, if something is missing. :smile:

tami5:
Wow @RishabhRD, support for Ultisnips is very important as many vim user depend on. Thanks

hrsh7th:
LGTM. Thank you ~.

PR #68, opened 2021-01-13
refresh completion when no new items are available

RishabhRD:
Hello @hrsh7th, This PR indents to refresh current completion even if no new
items are there for source. (i.e., when #source.items == 0)

This may not be the best way to achieve it, but it certainly would improve
behavior.

I would show it in gif.

Before:
before

After:
after

If you feel there is a better way to achieve it please tell me, I would change that in PR.:smiley:


hrsh7th:
It's my bad. LGTM!

PR #73, opened 2021-01-27
Remove &rtp search for ultisnips

comfortablynick:
Use `vim.F.npcall` to catch error instead, allowing ultisnips to be added to `&rtp` after nvim startup. Useful for those who do on-demand loading using built-in packages.

Also, the `match()` was reliant on the `ignorecase` setting; in my case (`noignorecase` set), it failed to match ultisnips even though it is installed and added via `packadd! ultisnips` in init.vim.

With this change, `vim.F.npcall()` will return `nil` if the call to `UltiSnips#{...}` fails -- meaning someone has added ultisnips as a completion source but doesn't have it loaded in nvim.


hrsh7th:
I've understand. Thank you.

Issue #74, opened 2021-01-28
How to bind cr work with Raimondi/delimitMate


glepnir:
As title Thanks:)

hrsh7th:
Probably, the below mappings will work.
```
  inoremap <silent><expr> <CR>      compe#confirm('<Plug>delimitMateCR')
```

glepnir:
Ok .found some thing wired with nvim-compe . Use lspsaga rename to rename a function then I enter insert mode, got failed.
`Not allowed here`, works fine with completion-nvim and disable nvim-compe.

hrsh7th:
`Not allowed here` relates to this code https://github.com/hrsh7th/nvim-compe/blob/master/lua/compe/completion.lua#L254

I don't know the actual cause please your guessing.

glepnir:
why this min config does not work what i am  missing?
```vim
set rtp+=~/.cache/vim/dein/repos/github.com/neovim/nvim-lspconfig
set rtp+=~/workstation/vim/lspsaga.nvim
set rtp+=~/.cache/vim/dein/repos/github.com/hrsh7th/nvim-compe

set completeopt=menu,menuone,noselect
lua <<EOF
require'lspconfig'.sumneko_lua.setup {
  cmd = {
    "/Users/stephen/workstation/lua-language-server/bin/macOS/lua-language-server",
    "-E",
    "/Users/stephen/workstation/lua-language-server/main.lua"
  };
  settings = {
    Lua = {
      diagnostics = {
        enable = true,
        globals = {"vim"}
      },
      runtime = {version = "LuaJIT"},
      workspace = {
        library = vim.list_extend({[vim.fn.expand("$VIMRUNTIME/lua")] = true},{}),
      },
    },
  }
}
require'compe'.setup {
  enabled = true;
  debug = false;
  min_length = 1;
  preselect = 'always';
  allow_prefix_unmatch = false;
  source = {
    nvim_lsp = true;
  };
}
EOF
```

glepnir:
solved.

glepnir:
hmm seems like <CR> does not work..
```
-- verbose inoremap <CR>
i  <CR>        * compe#confirm('<Plug>delimitMateCR')
```
then press cr I got `<Plug>delimitMateCR`

hrsh7th:
Oh sorry.
the `<Plug>delimitMateCR` should be expanded but the suggested mapping uses `inoremap`.

It seems to need to fix nvim-compe.

[...]

PR #80, opened 2021-01-28
Use feedkeys

### LINKED ISSUE #78: `compe#complete` should use `feedkeys` internally.

hrsh7th:
Fixes #78, #74

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,13 @@ inoremap <silent><expr> <CR>      compe#confirm(lexima#expand('<LT>CR>', 'i'))
 inoremap <silent><expr> <C-e>     compe#close('<C-e>')
 ```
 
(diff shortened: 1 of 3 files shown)


onsails:
when there is no completion candidates it fails with 
```
Error detected while processing function compe#confirm:
line    9:
E119: Not enough arguments for function: get
```

There is no such error with conditional map tho: `inoremap <silent><expr> <CR>      pumvisible() ? compe#confirm({ 'keys': "\<Plug>delimitMateCR", 'mode': '' }) : '<CR>'`

hrsh7th:
Thank you testing.
I've updated PR. Please re-check it.

onsails:
thank you for fixing!
It works correctly now.

hrsh7th:
Thank you for testing. I've merged.

Issue #83, opened 2021-01-28
Change to per-buffer configuration


comfortablynick:
I like the way this plugin works compared to completion-nvim, but it would be nice to be able to use it per buffer like most of the lsp-style plugins.

https://github.com/nvim-lua/completion-nvim/blob/fc9b2fd2d47bea6a8954de1b1b19f2330545b354/lua/completion.lua#L246-L265

This allows a lot of flexibility in how the plugin and associated autocmds are used. Currently, it looks like I'd have to do quite a bit of configuration just to prevent nvim-compe from triggering completion in places I don't want it (clap_input for example).

If nvim-compe provided an `on_attach` function, it could be used inside lsp `on_attach` or just a simple buffer autocmd. Is there any interest in this?

ii14:
#84

ii14:
I personally think it's better to do the deoplete approach, rather than what completion-nvim does. You provide the global config and then you can override the specific options for buffers.

It's easier to setup for users, it's trivial to implement in the current code base and you always have the option to make it act like `on_attach` function, by setting the global config to `{enabled = false}` and separately enabling it per buffer with desired options.

That's what's in my PR, although it assumes that the global config is defined at the moment, so it _might_ break if you don't call the `setup` function first. I'll be happy to further test it and fix the potential issues if that's the approach that @hrsh7th wants to take.

hrsh7th:
I want you to explain to me `why is this feature needed?`.

I've happy your guys contribution but I'm the careful man to add new structure for code base.

hrsh7th:
If the reason is this feature needed is `we want to stop auto-completation in specific buffer`, we should provide `let g:compe = { 'disabled' }` may be.

ii14:
I explained it in my PR:

> The use case is primarily different sets of sources for different filetypes.
> 
> ```viml
> let s:compe_c = {}
> let s:compe_c.source = {
>   \ 'nvim_lsp': v:true,
>   \ }
> au FileType c,cpp call compe#setup_buffer(s:compe_c)
> ```
> 
> For example, one might want to use the [necosyntax](https://github.com/tamago324/compe-necosyntax) source by default, but disable it when a language server is attached.
> 
> ```lua
> local lspconfig = require 'lspconfig'
> local compe = require 'compe'
> 
> lspconfig.clangd.setup{
>   on_attach = function()
>     compe.setup_buffer({
>       source = {
>         nvim_lsp = true,
>       },
>     })
>   end,
[...]

PR #84, opened 2021-01-28
Buffer local config

ii14:
#83 

The use case is primarily different sets of sources for different filetypes.

```vim
let s:compe_c = {}
let s:compe_c.source = {
  \ 'nvim_lsp': v:true,
  \ }
au FileType c,cpp call compe#setup_buffer(s:compe_c)
```

For example, one might want to use the [necosyntax](https://github.com/tamago324/compe-necosyntax) source by default, but disable it when a language server is attached.

```lua
local lspconfig = require 'lspconfig'
local compe = require 'compe'

lspconfig.clangd.setup{
  on_attach = function()
    compe.setup_buffer({
      source = {
        nvim_lsp = true,
      },
    })
  end,
}
```


comfortablynick:
Nice work, I think this works the way I was thinking. I had to remove all of the `if config.{field} ~= nil` guards and then it worked fine. The `config.{field} or {default}` already handles that, and the guard was keeping it from getting assigned at all.

After that, it seems to work well.

ii14:
> it assumes that the global config is defined at the moment, so it might break if you don't call the `setup` function first.

Maybe that was the problem?

My thought process behind the nil guards was that it should only override options that are defined in the `config` dictionary.

hrsh7th:
I will review it.

lua/compe/init.lua
```diff
@@ -38,7 +38,7 @@ local compe = {}
 compe.helper = Helper
 
 --- setup
-compe.setup = function(config)
+compe.setup = function(config, bufnr)
   Pattern.set_filetype_config('vim', {
```
ii14, line 42: `Pattern.set_filetype_config` can be potentially called multiple times. Is this a problem?

hrsh7th: Yes. We should move it another place.

autoload/compe.vim
```diff
@@ -5,15 +5,8 @@
...
-  call luaeval('require"compe".setup(_A[1])', [a:config])
-endfunction
-
-"
-" compe#setup_buffer
-"
-function! compe#setup_buffer(config) abort
-  call luaeval('require"compe".setup_buffer(_A[1])', [a:config])
+function! compe#setup(config, ...) abort
+  call luaeval('require"compe".setup(_A[1], _A[2])', [a:config, get(a:, 1, v:null)])
```
ii14, line 9: Should we validate the arguments and throw errors when, for example, wrong type is passed or there are more arguments than expected?

hrsh7th: No. The current implementation is enough 👍

lua/compe/init.lua
ii14: Maybe on initialization, in `plugin/compe.vim`?

hrsh7th: I think the top of `init.lua` would be good.

lua/compe/config.lua
```diff
@@ -35,50 +35,50 @@ Config._normalize = function(config)
...
-  Config._config = Config._normalize(config)
-end
-
-Config.setup_buffer = function(config)
-  local buf_config = Config._normalize(vim.deepcopy(Config._config))
+Config.setup = function(config, bufnr)
+  if bufnr == nil then
+    Config._config = Config._normalize(config)
+  else
+    local buf_config = Config._normalize(vim.deepcopy(Config._config))
```
hrsh7th, line 42: I think it should be more simply.

```lua
config = Config._normalize(config)

if bufnr == nil then
  Config._config = config
else
  for key, value in pairs(Config._config) do
    if config[key] == nil then
      config[key] = value
    end
  end
  Config._bufnrs[bufnr] = config -- or vim.api.nvim_buf_set_var(bufnr, 'compe_config', config)
end

```

hrsh7th: I think `vim.api.nvim_buf_set_var` is not required for it... maybe.

hrsh7th: Probably, the merging logic should be moved to `Config.get`, I think.

ii14: Much nicer. :+1:

ii14: > Probably, the merging logic should be moved to `Config.get`, I think.

Doing it in `Config.setup` is a minor optimization, it will be evaluated only once. But I don't know how often `Config.get` is called, so maybe it doesn't really matter.

hrsh7th: I think we don't worry to performance for it.
If this worse to performance, we can use `utils.cache` module (with versioning).

lua/compe/init.lua
```diff
@@ -32,22 +32,23 @@ end
...
+Pattern.set_filetype_config('vim', {
+  keyword_pattern = [[\%(\h\%(\w\|#\)*\)]];
+})
+Pattern.set_filetype_config('php', {
+  keyword_pattern = [[\%(\$\w*\|\h\w*\)]];
+})
+Pattern.set_filetype_config('html', {
+  keyword_pattern = [[\%(/\h\?\w*\|\h\w*\)]];
+})
+
```
hrsh7th, line 44: 👍

lua/compe/config.lua
```diff
@@ -20,27 +22,51 @@ Config.setup = function(config)
...
+  config = Config._normalize(config)
+
+  if bufnr == nil then
+    Config._config = config
+  else
+    for key, value in pairs(Config._config) do
+      if config[key] == nil then
+        config[key] = value
+      end
+    end
```
hrsh7th, line 50: It is a shallow merge.

So suggested use-case will be realized by the below codes.

```viml
lua <<EOF
local compe = require"compe"

-- global configuration.
compe.setup {
  source = {
    necosyntax = true,
    buffer = true,
    path = true,
  }
}

local on_attach = function()
  -- Disable necosyntax source when nvim_lsp is activated.
  compe.setup({
    source = {
      nvim_lsp = true,
      buffer = true,
      path = true,
    }
  }, 0)
end
require"lspconfig".vimls.setup {
  on_attach = on_attach
}
EOF
```

Is it enough for @ii14?

ii14: Almost, the merging doesn't work. I'm trying to figure out what's wrong.

ii14:
Everything seems to work now.

hrsh7th:
LGTM.
I will try it on my local machine then I will merge it.

hrsh7th:
It works great.
I want to add the documentation for it. Can you please?

I try to realize disable buffer completion for the buffer that enabled nvim_lsp, it can be done straightforward.

```
  local on_attach = function()
    require'compe'.setup({
      source = {
        nvim_lsp = true;
        vsnip = true;
        path = true;
      }
    }, 0)
  end
```

ii14:
> I want to add the documentation for it. Can you please?

Sure, no problem.

doc/compe.txt
```diff
@@ -298,12 +298,12 @@ DEVELOPMENT                                                          *compe-dev*
 ==============================================================================
 EXAMPLE SOURCE                                             *compe-dev-ex-source*
 
-You can see example on *https://github.com/kristijanhusak/vim-dadbod-completion*
```
ii14, line 301: @elianiva `**` will always create a new tag, so this link was showing up in `:help` completion on the command line. Just to let you know :)

ii14:
I'll try to work on the vim docs in a separate PR, but for now this should be enough.

hrsh7th:
Thank you very much. You did great work.
The docs can be improved later if need. In my view, these changes are good.

doc/compe.txt
elianiva: thanks!

PR #89, opened 2021-01-29
docs: improved docs

elianiva:
I met some people that get confused by nvim-compe docs so I tried to "improve" the docs so people can get started more easily. Some things might be missing like `allow_prefix_unmatch` because I don't know what this option does and maybe some other explanations.

I was thinking of adding how to use this with other autopairs such as `nvim-autopairs` because you have an example for `lexima` but I'm not sure if that's the scope of this project.

edit: I can also help converting the readme into vimdocs if you want :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,81 +2,134 @@
 
 Auto completion plugin for nvim.
 
+## Table Of Contents
+- [Concept](#concept)
+- [Usage](#usage)
+  - [Prerequisite](#prerequisite)
+  - [Configuring in Vimscript](#configuring-in-vimscript)
+  - [Configuring in Lua](#configuring-in-lua)
+  - [Mappings](#mappings)
+  - [Source Configuration](#source-configuration)
+- [Builtin Sources](#builtin-sources)
+- [Development](#development)
+  - [Example Source](#example-source)
+  - [The Source](#the-source)
+  - [Public API](#public-api)
+    - [Vimscript](#vim-script)
+    - [Lua](#lua)
 
-# Concept
+
+## Concept
 
 - Lua source & Vim source
 - Better matching algorithm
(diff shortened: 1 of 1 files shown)


hrsh7th:
Realy helpful. Thank you for @elianiva

README.md
```diff
@@ -2,81 +2,134 @@
...
+let g:compe.source.path = v:true " path completion. default: `false`
+let g:compe.source.buffer = v:true " buffer completion. default: `false`
+let g:compe.source.vsnip = v:true " vsnip completion, make sure you have `vim-vsnip` installed. default: `false`
+let g:compe.source.nvim_lsp = v:true " nvim builtin LSP completion. default: `false`
+let g:compe.source.nvim_lua = v:true " nvim's lua stdlib completion. default: `false`
+let g:compe.source.your_awesome_source = {
+  " you can also override completion source configuration
+  " by setting it to a `table` instead of `boolean`
+  " see `Source Configuration` for more information
+  \}
```
hrsh7th, line 76: Sorry. I don't want to maintein same explanation with Lua example and VimL example both.
Can you remove this duplication?

elianiva: sure

hrsh7th: It would be good to add a separate section. Thank you.
I don't stick to comment for explanation.

hrsh7th:
LGTM! Thank you very much.

If you have no more improvements in this PR, I will merge it.

elianiva:
Yep, I think I'm done.

Probably want to work on vimdocs version of this so people can just use `:h nvim-compe` if they want to look the documentation, but that's a separate PR because that might take some time.

hrsh7th:
Agree with you. I skipped vim's docs...

Issue #85, opened 2021-01-28
No dictionary auto-completion?


coxackie:
I may be missing this entirely, but I have not seen dictionary source, and I can confirm that auto-completion is not using dictionary. This is quite an important source for an editor; as I said, i may be missing this altogether. Please advise.

elianiva:
I was also thinking about this last night, being able to auto-complete an actual word and seeing its definition is great. But I'm worried about how big the source would be, [this repo](https://github.com/matthewreagan/WebstersEnglishDictionary) has an English Dictionary in form of JSON that we *could* use but it's huge, around 20MB.

hrsh7th:
I was misunderstanding. vim's `getcompletion` has no returns `dict` completion results.

coxackie:
I think I also did not explain well. I meant the simple spell suggestions one would get spell check on. The ones with C-n or C-p. This should be possible with no overhead in performance, right?

hrsh7th:
If you want to realize vim's native spell completion. You can look `compe_tags` source as reference implementations.

elianiva:
Ah! I forgot if (n)vim has dictionary for completion already.

hrsh7th:
we can use `spellsuggest({word})` ...? I haven't ever used it.

coxackie:
Completion.nvim has something called "modes completion", where you set basically the keyboard trigger from omnicomplete and it triggers it. So you literally just add as a mode the trigger keystroke.

hrsh7th:
I absolutely never implement `modes completion`. sorry.

coxackie:
It is your project and you do what you want. And thank you very.much. If there is another way of doing this and you can, I would be happy and grateful.

hrsh7th:
The spell source has merged. Feel free to close this issue.

I think we can improve spell source by parsing spell files manually by Lua.
I'm looking forward to PR from someone.

coxackie:
OK, many thanks!

 I am going to close this, but maybe you want to open a new issue with your desire for spell source by parsing spell files manually by Lua?

PR #91, opened 2021-01-29
spell source

hrsh7th:
https://github.com/hrsh7th/nvim-compe/issues/85

@coxackie What do you think about it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,16 +150,23 @@ The sources can be configured by `let g:compe.source['source_name'] = { ...confi
 
 ## Built-in sources
 
+#### common
+
 - buffer
 - path
 - tags
+- spell
+
+#### neovim-specific
+
 - nvim_lsp
 - nvim_lua
+
+#### external-plugin
+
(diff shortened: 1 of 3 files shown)


coxackie:
I can confirm it works; I set `spell = true;` to sources, and it looks quite OK. I will continue using it and will report if anything weird happens, but maybe you can merge and update readme, so you get more feedback.

 One remark; suggestions come with [Spell] as indicator; others such as snippets, buffer come like [VSNIP], [BUFFER]. I personally like better the smallcap; they could all be made like [VSnip], [Buffer]. Capital letters may be too much. I do not want to open a full issue on this, but I can do if you think it should be discussed.

 Great work - many thanks!

cloggier:
I agree with all uppercase being too busy. I was thinking of changing them all locally, but would prefer if that was changed upstream.

hrsh7th:
The PR is welcome. I don't stick to uppercase letter (just dragged to `LSP`)

PR #93, opened 2021-01-29
Add option to disable autocomplete

### LINKED ISSUE #88: Disable auto popup
Is it possible to disable automatically open the completion popup?

The plugin works really well. But I like to trigger completion only manually when I need it.

lukas-reineke:
if `let g:compe.autocomplete` is set to false, completion only triggers manually

fix #88

Let me know if you want to change the name to something better.
`completion-nvim` calls it `completion_enable_auto_popup`.

### DIFF
--- a/lua/compe/config.lua
+++ b/lua/compe/config.lua
@@ -18,6 +18,11 @@ Config.setup = function(config)
   config.source_timeout = config.source_timeout or SOURCE_TIMEOUT
   config.incomplete_delay = config.incomplete_delay or INCOMPLETE_DELAY
   config.allow_prefix_unmatch = Config._true(config.allow_prefix_unmatch)
+  if config.autocomplete == nil then
+    config.autocomplete = true
+  else
+    config.autocomplete = Config._true(config.autocomplete)
+  end
 
   -- normalize source metadata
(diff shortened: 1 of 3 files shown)


hrsh7th:
Looks good.
Could you add documentation for this option?

lua/compe/config.lua
```diff
@@ -18,6 +18,10 @@ Config.setup = function(config)
   config.source_timeout = config.source_timeout or SOURCE_TIMEOUT
   config.incomplete_delay = config.incomplete_delay or INCOMPLETE_DELAY
   config.allow_prefix_unmatch = Config._true(config.allow_prefix_unmatch)
+  config.autocomplete = Config._true(config.autocomplete)
```
hrsh7th, line 21: `_true` will normalize `nil` as `false`.
So I think the below condition always returns false, right?

lukas-reineke: yeah you're right..

pushed a fix

lukas-reineke: or maybe `_false` function that defaults to true on anything other than `true` or `0` would be cleaner?

hrsh7th:
LGTM! Thank you for your contribution!

PR #94, opened 2021-01-29
docs: added vimdocs

elianiva:
Follow up of #89. This is my first time writing vim docs so let me know what you think :)

edit: I also made the `enabled` field to true by default based on #95 so users don't have to explicitly enable it.


hrsh7th:
Thank you for awesome contribution!

doc/nvim-compe.txt
```diff
@@ -0,0 +1,352 @@
+*nvim-compe* *compe*
```
hrsh7th, line 1: I prefer `compe.txt` as filename.

hrsh7th:
LGTM.
I will merge it after filename changed.

doc/nvim-compe.txt
elianiva: done

hrsh7th:
Thank you!

PR #108, opened 2021-01-31
Add Ultisnips Confirm Function

srafi1:
Add a confirm function for the ultisnips completion source. Ultisnips normally has its own expand trigger, but mapping it to `<CR>` can be problematic as it interferes with a lot of other plugins that try to map the same key. When used in conjunction with such plugins, I think it is better to defer to nvim-compe's completion.


hrsh7th:
Thank you. It's useful.

PR #116, opened 2021-02-02
Fixed setting enabled to false

### LINKED ISSUE #114: can't disable nvim-compe
```vim
augroup Compe
  autocmd!
  autocmd BufEnter * let g:compe_enabled = v:true
  autocmd FileType clap_input let g:compe_enabled = v:false
augroup END
```
setting `g:compe_enabled` doesn't actually disable compe, and setting `enabled = false` in the setup also doesn't disable completion

ii14:
Fixes #114

### DIFF
--- a/lua/compe/config.lua
+++ b/lua/compe/config.lua
@@ -12,7 +12,11 @@ Config._bufnrs = {}
 
 Config._normalize = function(config)
   -- normalize options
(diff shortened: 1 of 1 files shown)


hrsh7th:
It's good for now.

I think we should remove `_true(value)` but add `_bool(value, [default])` instead.

hrsh7th:
Thank you!

Issue #120, opened 2021-02-02
ideas


hrsh7th:
Feel free to post ideas. (I don't know if i do)

hrsh7th:
git source.

tami5:
> git source.

hmm interesting

hrsh7th:
I don't know that is this source if teally useful.
But maybe all branch name or log (with docemtation)

hrsh7th:
project files source. (provodes all relative file paths)

hrsh7th:
register or yank history source

ii14:
> project files source. (provodes all relative file paths)

Maybe there should be an option for `path` source to include files from the current git repo?

ii14:
The [necosyntax](https://github.com/tamago324/compe-necosyntax) is really cool, but it has a dependency and I noticed that it slows down vim startup time. Maybe rewriting it to lua and somehow doing it asynchronously would help? I'll try to look into that when I have some free time.

cloggier:
> register or yank history source

Getting yanks from miniyank perhaps?

cloggier:
How about lbdb source to complete email addresses from address book while composing emails?

ii14:
> The necosyntax is really cool, but it has a dependency and I noticed that it slows down vim startup time. Maybe rewriting it to lua and somehow doing it asynchronously would help? I'll try to look into that when I have some free time.

Nope, I got it wrong, just checked it again and it works just fine. Sorry.

ii14:
@hrsh7th I know that you've said you don't like completions from other buffers, but some people like to have that so maybe it makes sense to provide that?

hrsh7th:
I don't want to provide settings as much as possible.

In my random thoughts, we provide specific prefix 
for triggering some source.

For example,
1. !/ -> project files
2. @. -> yank-history

But I don't know if these useful 😅

hrsh7th:
Probably, we shoyld design source option interface.

ii14:
I don't know if it would be actually useful for anyone, but what about `compe#completion()` optionally taking in source as argument?
You could for example create mappings like builtin `<C-X><C-F>`:
```vim
inoremap <silent><expr> <C-X><C-F> compe#complete('path')
```

hrsh7th:
It' good idea.
I will consider it.

tamago324:
> git source.

What is git source, and is it something like https://github.com/zk-phi/git-complete?
[...]

PR #128, opened 2021-02-03
Add LSC source

hrsh7th:
If it is not enough, please create PR (because I don't use vim-lsc now).

Related https://github.com/hrsh7th/nvim-compe/issues/120#issuecomment-772225612

### DIFF
--- a/after/plugin/compe_vim_lsc.vim
+++ b/after/plugin/compe_vim_lsc.vim
@@ -1,0 +1,11 @@
+if exists('g:loaded_compe_vim_lsc')
+  finish
+endif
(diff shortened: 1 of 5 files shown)


bluz71:
Hello @hrsh7th,

This PR fails to run for me due to the following error:

```sh
Error detected while processing /home/db/dotfiles/vim/plugged/nvim-compe/autoload/compe_vim_lsc/source.vim:
line    1:
E117: Unknown function: vital#lamp#import
E15: Invalid expression: vital#lamp#import('VS.LSP.MarkupContent')
```

I would like to test vim-lsc integration but this error is preventing that.

Hopefully it is an easy fix.

PR #131, opened 2021-02-03
refactor: move vim.fn to vim.api whenever possible

elianiva:
use `vim.api` to replace `vim.fn` when possible, the performance should increase slightly, I think.
I did this because I saw this chat on Neovim's Gitter.

I haven't tested it thoroughly cause I don't have enough time atm, but from what I experienced, nothing's broken.

image


hrsh7th:
That's make sense.

Thank you very much!

PR #133, opened 2021-02-04
feat: added treesitter source

elianiva:
I don't use it personally but if people coming from completion.nvim want to have this source on nvim-compe as well, might worth to merge this in. Otherwise, feel free to close this PR if no one wants it.

### DIFF
--- a/after/plugin/compe_treesitter.vim
+++ b/after/plugin/compe_treesitter.vim
@@ -1,0 +1,9 @@
+if exists('g:loaded_compe_treesitter')
+  finish
+endif
+let g:loaded_compe_treesitter = v:true
+
+if exists('g:loaded_compe')
+  lua require'compe'.register_source('treesitter', require'compe_treesitter')
+endif
+
--- a/lua/compe_treesitter/init.lua
+++ b/lua/compe_treesitter/init.lua
@@ -1,7 +1,7 @@
 local ts_exists = pcall(require, "nvim-treesitter")
-local ts_locals = require('nvim-treesitter.locals')
-local parsers = require('nvim-treesitter.parsers')
-local ts_utils = require('nvim-treesitter.ts_utils')
+local _, ts_locals = pcall(require, 'nvim-treesitter.locals')
+local _, parsers = pcall(require, 'nvim-treesitter.parsers')
+local _, ts_utils = pcall(require, 'nvim-treesitter.ts_utils')
 local compe = require('compe')
 
 local Source = {}
--- a/lua/compe_treesitter/init.lua
+++ b/lua/compe_treesitter/init.lua
@@ -1,0 +1,87 @@
+local ts_locals = require('nvim-treesitter.locals')
+local parsers = require('nvim-treesitter.parsers')
+local ts_utils = require('nvim-treesitter.ts_utils')
+local compe = require('compe')
+
+local Source = {}
+
+function Source.new()
+  return setmetatable({}, { __index = Source })
+end
+
+function Source.get_metadata(_)
+  return {
+    priority = 100;
+    dup = true;
+    menu = '[Treesitter]';
+  }
+end
+
+function Source.determine(_, context)
+  return compe.helper.determine(context)
+end
+
+function Source._get_smallest_context(_, source)
+  local scopes = ts_locals.get_scopes()
+  local current = source
+
+  while current ~= nil and not vim.tbl_contains(scopes, current) do
+    current = current:parent()
+  end
+
+  return current or nil
+end
+
+function Source._prepare_match(self, match, kind)
+  local matches = {}
+
+  if match.node then
+    table.insert(matches, { kind = kind or '', node = match.node })
+  else
+    for name, item in pairs(match) do
+      vim.list_extend(matches, self:_prepare_match(item, name))
+    end
+  end
+
+  return matches
+end
+
+function Source._get_treesitter_items(self)
+  if parsers.has_parser() then
+    local complete_items = {}
+
+    local at_point = ts_utils.get_node_at_cursor()
+    local line_current = vim.api.nvim_win_get_cursor(0)[1]
+
+    for name, definitions in ipairs(ts_locals.get_definitions(0)) do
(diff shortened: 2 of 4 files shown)


weilbith:
Sorry, this is not really related to this PR, but it popped up in my notifications and you might be able to give some insights.
What I'm currently missing and `completion.nvim` is the feature of independent asynchronous sources. So I have some big projects where the language server slows down. What I would like to have is tree-sitter as a fast completion source with low priority level and native LSP with high priority. So in case that LSP is slow, I can still get fast on-the-fly completion suggestions (e.g. long variable names) via Treesitter. But when LSP resolves its candidate, they get added to the list and take higher priority in the sorting order than the Treesitter once as I argue they are more accurate. An example would be the trying of a class instance variable and then waiting for the completion of possible methods of the class or its properties etc. Just because you don't remember the library anymore or what ever.
Is this kinda possible with this completion plugin?

elianiva:
You can already do that, I think. Just make sure treesitter source has lower priority than LSP source (which I already did if I remember correctly, so no configuration needed)

cloggier:
It would be nice if it then removed lower priority items if they are duplicates. Just a thought.

hrsh7th:
The all of above discussion is the same to current behavior.

weilbith:
Wow, cool this sounds awesome.
So just completion sources are treated independent in this plugin and the completion menu gets populated when the first source responds and then gets continuously extended each time another source responds?

weilbith:
> The all of above discussion is the same to current behavior.

As I stated this was not directly targeting this PR but being a question about the plugin.

hrsh7th:
Almost yes, But sometimes it makes flicker so this plugin has `source_timeout` option.

weilbith:
Cool. If this is possible you got me and I'll switch other to this plugin soon. This is what I was awaiting for. :hugs: :partying_face:

hrsh7th:
I will review it after experiencing it.

hrsh7th:
I've tried it. It is useful than my thoughts.

after/plugin/compe_treesitter.vim
```diff
@@ -0,0 +1,9 @@
+if exists('g:loaded_compe_treesitter')
+  finish
+endif
+let g:loaded_compe_treesitter = v:true
+
+if exists('g:loaded_compe')
```
hrsh7th, line 6: we should check `nvim-treesitter` is installed or not.
(You can check it in here or compe_treesitter/init.lua)

lua/compe_treesitter/init.lua
```diff
@@ -0,0 +1,87 @@
...
+      for _, match in ipairs(matches) do
+        local node = match.node
+        local node_scope = self:_get_smallest_context(node)
+        local start_line_node, _, _= node:start()
+        local node_text = ts_utils.get_node_text(node, 0)[1]
+
+        if node_text
+          and (not node_scope or ts_utils.is_parent(node_scope, at_point))
+          and (start_line_node <= line_current) then
+          table.insert(complete_items, {word = node_text, kind = match.kind})
```
hrsh7th, line 68: The kind is number so it is not useful... Hm... we should define the mapping?

lua/compe_treesitter/init.lua
```diff
@@ -0,0 +1,87 @@
...
+    for name, item in pairs(match) do
+      vim.list_extend(matches, self:_prepare_match(item, name))
+    end
+  end
+
+  return matches
+end
+
+function Source._get_treesitter_items(self)
+  if parsers.has_parser() then
```
hrsh7th, line 50: This check should move to `determine`.

lua/compe_treesitter/init.lua
```diff
@@ -0,0 +1,87 @@
...
+    return complete_items
+  end
+
+  return {}
+end
+
+function Source.complete(self, context)
+  context.callback({
+    items = self:_get_treesitter_items(),
+    incomplete = true
```
hrsh7th, line 82: I think it is not necessary.

lua/compe_treesitter/init.lua
elianiva: I agree, it is not *that* useful at the moment

elianiva:
Thanks for the review! I'll fix them tomorrow since I won't be able to access my laptop today.

lua/compe_treesitter/init.lua
elianiva: do  you still see the kind field filled with number? I see them as `var`, `import`, `associated`, etc.

hrsh7th: I can't see the number at the moment. Thank you.

lua/compe_treesitter/init.lua
```diff
@@ -0,0 +1,87 @@
+local ts_exists = pcall(require, "nvim-treesitter")
```
hrsh7th, line 1: I think this is not enough because the still require `nvim-treesitter.locals` without pcall.

elianiva: I think if someone have nvim-treesitter installed, they will always have `nvim-treesitter.locals` so I assumed we only need to check the top module.

hrsh7th: If the user does not have `nvim-treesitter`, `require'nvim-treesitter.locals` will throw exception.

1. `local ts_exists = pcall(require, "nvim-treesitter")` ignore exception.
2. `local ts_locals = require('nvim-treesitter.locals')`

Currently, nvim-compe uses `pcall` for public apis.

Could you check with `let g:compe.debug = v:true`? (If you have a time)

elianiva: ah, yeah, that makes sense. Will fix it in a moment

hrsh7th: Good! Thank you!

hrsh7th:
It almost looks good to me.

I'm merging it. if someone wants to improve it more, please create PR.

hrsh7th:
@elianiva Thank you very much! I'm happy for your contribution.

PR #136, opened 2021-02-04
fix: avoid creating swapfiles for preview buffers

nanotee:
Sometimes Neovim fails to delete swapfiles which can result in repeated error messages upon re-opening the editor

Thanks for maintaining this really nice plugin!


hrsh7th:
LGTM! Thank you!

cloggier:
So this is what was happening! :) Thanks @nanotee!

PR #137, opened 2021-02-05
documentation: add window var to identify

fsouza:
This is useful for applying themes to the window (using capabilities in
neovim 0.5).


hrsh7th:
lt's the correct solution. Thank you.

autoload/compe/documentation.vim
```diff
@@ -7,6 +7,7 @@ let s:FloatingWindow = vital#compe#import('VS.Vim.Window.FloatingWindow')
 let s:window = s:FloatingWindow.new()
 call s:window.set_var('&wrap', 1)
 call s:window.set_var('&conceallevel', 2)
+call s:window.set_var('nvim_compe_documentation', 1)
```
hrsh7th, line 10: I think `nvim_` prefix should be removed.

fsouza: Thank you for the quick review!

hrsh7th:
LGTM. Thank you very much!

PR #134, opened 2021-02-04
Add sticky_char (experimental)

hrsh7th:
see https://github.com/hrsh7th/nvim-compe/issues/120#issuecomment-773165747

### TODO
- [ ] Consideration
- [ ] Improve determine with sticky_char

PR #139, opened 2021-02-05
RFC: `compe#reduce` mapping

hrsh7th:
This is another idea of https://github.com/hrsh7th/nvim-compe/pull/134 .

### DIFF
--- a/autoload/compe.vim
+++ b/autoload/compe.vim
@@ -70,6 +70,14 @@ function! compe#confirm(...) abort
 endfunction
 
 "
+" compe#reduce
+"
+function! compe#reduce() abort
+  call luaeval('require"compe"._reduce()')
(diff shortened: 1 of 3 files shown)


hrsh7th:
I think this is the ideal solution for https://github.com/hrsh7th/nvim-compe/issues/120#issuecomment-773165747

lucax88x:
I would like to try this, can you give some example configuration?

elianiva:
> I would like to try this, can you give some example configuration?

```lua
vim.api.nvim_set_keymap('i', '<A-r>', 'compe#reduce()', { noremap = true, expr = true, silent = true })
```

press Alt + r (you can change this of course) when the completion menu appears, it should remove one of the source and you'll get less completion items.

PR #144, opened 2021-02-06
Add new option to change documentation behaviour

### LINKED ISSUE #98: Is there any way to disable preview on completion?
Title.

sblumentritt:
This resolves #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ let g:compe.incomplete_delay = 400
 let g:compe.max_abbr_width = 100
 let g:compe.max_kind_width = 100
 let g:compe.max_menu_width = 100
+let g:compe.documentation = v:true
 
 let g:compe.source = {}
 let g:compe.source.path = v:true
@@ -123,6 +124,7 @@ require'compe'.setup {
   max_abbr_width = 100;
   max_kind_width = 100;
   max_menu_width = 100;
+  documentation = true;
 
   source = {
     path = true;
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -91,10 +91,12 @@ Completion.select = function(index)
   if completed_item then
     Completion._selected_item = completed_item
 
-    for _, source in ipairs(Completion.get_sources()) do
-      if source.id == completed_item.source_id then
-        source:documentation(completed_item)
-        break
+    if Config.get().documentation then
+      for _, source in ipairs(Completion.get_sources()) do
+        if source.id == completed_item.source_id then
+          source:documentation(completed_item)
+          break
+        end
       end
     end
   end
--- a/doc/compe.txt
+++ b/doc/compe.txt
@@ -269,6 +269,11 @@ max_menu_width ~
     Type: |Number|
     Default: 100
 
+documentation ~
+    Documentation behaviour.
+      Type: |Boolean|
+      Default: true
+
 source ~
     Source configuration. Required.
       Type: |Dictionary| of:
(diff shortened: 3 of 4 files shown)


elianiva:
wouldn't it be better as `show_documentation` ? Because nvim-compe only shows the documentation on a floating window anyway.

hrsh7th:
Yes. I think option name should be changed.
I prefer `documentation = 'enable' || 'disable'`.

lua/compe/completion.lua
```diff
@@ -95,18 +95,20 @@ end
 
 --- select
 Completion.select = function(index)
-  if index == -1 then
-    return
-  end
+  if Config.get().show_documentation_as_float then
+    if index == -1 then
```
hrsh7th, line 99: This check should be moved to L107.

We must set `_selected_item` even if documentation disabled.

cloggier:
This. There might be future options added later and then having `documentation` could become a table with more specific options.

sblumentritt:
I would prefer the `show_documentation` with `true || false` as boolean checks are more performant than string checks.

hrsh7th:
We must ignore minor optimization in design decisions, I think.

If the user wants to disable documentation only when `preselect`, the boolean flags can't extend for it.

lua/compe/completion.lua
```diff
@@ -103,10 +103,12 @@ Completion.select = function(index)
   if completed_item then
     Completion._selected_item = completed_item
 
-    for _, source in ipairs(Completion.get_sources()) do
-      if source.id == completed_item.source_id then
-        source:documentation(completed_item)
-        break
+    if Config.get().documentation == 'enable' then
```
hrsh7th, line 106: Thank you! It's correct!

cloggier:
I am with @sblumentritt here. Should be bool to start with. More detailed configuration can be part of table this option could accept.

hrsh7th:
LGTM!
Thank you for your contribution! 😄

hrsh7th:
@cloggier 
Is this what you are mean?

```
let g:compe = {}
let g:compe.documentation = v:true || {
\   'show': 'preselect',
\   'layout': 'docked',
\ }
```

hrsh7th:
```
let g:compe = {}
let g:compe.documentation = v:true
```

```
let g:compe = {}
let g:compe.documentation = {}
let g:compe.documentation.disabled = v:true
```

cloggier:
Yes, something like that. Same as sources can be true or a table for more specific configuration.

hrsh7th:
You are absolutely right. I'm wrong. sorry!!

@sblumentritt Really sorry! I had the wrong opinion. Could you change to the above format?

sblumentritt:
No problem 😄 

Do you want the `disabled` as boolean entry or should we use the `show` entry with `'enable' || 'disable'` as string entry?

hrsh7th:
Currently, it is ok to be just string flag.

README.md
```diff
@@ -123,6 +126,10 @@ require'compe'.setup {
   max_kind_width = 100;
   max_menu_width = 100;
 
+  documentation = {
+    show = 'enable';
+  };
```
hrsh7th, line 131: Sorry for my poor English.

My intention is we should provide only `let g:compe.documentation = v:true | v:false` for now.

hrsh7th:
LGTM. Sorry for the confusion.
Thank you very much!

bluz71:
Many thanks to those you implemented and reviewed this.

This is a very useful option since some of us are minimalists.

nvim-compe keeps getting better.

PR #149, opened 2021-02-07
Add TabNine Souce

tzachar:
Initial proposal for TabNine source.
Mostly operational.

The only thing remaining is exposing some configuration variables to users.

Hope you can merge.


install_tabnine.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
ii14, line 1: Assuming we should even provide the installation script in the first place, it definitely shouldn't be in the root directory. At least in my opinion. For now the best place for it would be in the `misc` directory.

hrsh7th:
I will look into it. Thank you.

lua/compe_tabnine/init.lua
```diff
@@ -0,0 +1,182 @@
...
+	local cur_line_after = string.sub(cur_line, cursor[2]+1) -- include current character
+
+	local region_includes_beginning = false
+	local region_includes_end = false
+	if cursor[1] - Source.max_lines <= 1 then region_includes_beginning = true end
+	if cursor[1] + Source.max_lines >= fn['line']('$') then region_includes_end = true end
+
+	local lines_before = api.nvim_buf_get_lines(0, cursor[1] - Source.max_lines , cursor[1]-1, false)
+	table.insert(lines_before, cur_line_before)
+	local before = fn.join(lines_before, "\n")
```
elianiva, line 76: I think this could be replaced with `table.concat(lines_before, "\n")`, we should try not to use `vim.fn` that often because that might cause some performance issue.

tzachar: Sure.
Any other Lua fixes will be appreciated as Im not that fluent in Lua.

misc/install_tabnine.sh
```diff
@@ -0,0 +1,34 @@
...
+path=$version/$triple/TabNine
+if [[ -f "binaries/$path" ]] && [[ -e "binaries/TabNine_$(uname -s)" ]]; then
+    ln -sf $path "binaries/TabNine_$(uname -s)"
+    exit
+fi
+echo Downloading version $version
+curl https://update.tabnine.com/$path --create-dirs -o binaries/$path
+chmod +x binaries/$path
+
+ln -sf $path "binaries/TabNine_$(uname -s)"
```
hrsh7th, line 34: Sorry. I want to avoid adding the ability to install some binaries on nvim-compe.

So the `tabnine` source should check `tabnine` command in $PATH.

elianiva: agree, maybe we could add this script to the wiki if people want to know how to install it

tzachar: Well, this is your project after all.. However, I do believe having 'batteries included' is better. After all, users have the option to install the binary if they so chose when they load the plugin.

If we leave it the the user, its one more dependency for them to track manually. Its much easier when the plugin manager takes care of that.

Can you elaborate on you opposition to adding the install script?

onsails: In some environments installing binary on plugin activation is not an option, for example on [nixos](https://nixos.org/) or other os with nix package manager maaging vim plugins. [Here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/misc/vim-plugins/overrides.nix#L744) you can see how completion-tabnine has nothing to do with tabnine install from inside of the plugin, it just depends on the separate tabnine package.

tzachar: Well, the actual script was taken from aca/completion-tabnine, so im not sure I follow your logic.
Anyway, we can have a slightly different strategy of locating the tabnine binary:
1) try 'TabNine' from $PATH
2) if (1) fails, try the current approach (which would work if the user opted to call the install script)
3) if all fails, show an appropriate message.

Would this work?

onsails: I've just noticed that you suggest just an optional tabnine install. That will work for nix because in nix we just don't execute such install scripts (they are considered non-pure and non-reproducible).

hrsh7th: For example, if we provide `github completion` for `COMMIT_EDITMSG`, Should we add `gh` command install scripts?
And In this case, should we suggest configuration as below?

```viml
Plug 'hrsh7th/nvim-compe', { 'do': './misc/install_tabnine.sh | ./misc/install_gh.sh' }
```

I think this is over responsibility. (And Sorry. I don't well shellscript so I can't maintenance it).

tzachar: I understand you not wanting to maintain it. Its a valid consideration.

However, gh is more common, and much easier to install. TabNine is not and, moreover (alas), propietary. Every other instance of a plugin using TabNine (completeion-nvim, deoplete, and others) do offer the install script. The only difference is that the plugin is not maintained in the main repo, but as a separate plug. 

Would you prefere I create a separate repo for compe-tabnine? 
If so, can you please explain how to register the source in compe?

tzachar:
Anyway, I am having an issue with the current design.
I expect that if Source.complete calls the callback like this:
```lua
  args.callback({               
          items = nil;          
          incomplete = true;    
  })                            
```
there will be another call to Source.complete in the near future to get the results.
I am probably missing something. 

The design probelm is that in the first call to Source.complete I initiate a request to the TabNine process, which will be done asynchronously, and the results will only be available in a future call to Source.complete.
Is there any way to support this behaviour?

misc/install_tabnine.sh
hrsh7th: Thank you for your words.

Yes. If we should provide installation scripts, the source should be separated repository (I think `compe-tabnine` is good).

The source registration is super simple and easy. You can create below structure.

```sh
/compe-tabnine
  /after
    /plugin
      compe_tabnine.vim
  /lua
    /compe_tabnine
      init.lua
```

tzachar: ok, I will switch to a seperate repo.
would appreciate if you can reference it in you documentation.

hrsh7th:
You can call `args.callback` asynchronously. So you can wait tabnine response.

hrsh7th:
For example, the nvim_lsp source will receive items in asynchronously.
https://github.com/hrsh7th/nvim-compe/blob/master/lua/compe_nvim_lsp/source.lua#L46

tzachar:
ok, moved to tzachar/compe-tabnine

PR #150, opened 2021-02-08
add check_back_space in tab

glepnir:


hrsh7th:
LGTM. Thank you.

README.md
```diff
@@ -231,6 +231,15 @@ local t = function(str)
   return vim.api.nvim_replace_termcodes(str, true, true, true)
 end
 
+local check_back_space()
```
cloggier, line 234: That function definition doesn't seem right, should it not be
```lua
local check_back_space = function()
```

glepnir: yep. ... I delete _g miss funtion

PR #154, opened 2021-02-08
fix miss function keyword

glepnir:


hrsh7th:
Sorry. I couldn't notice it. Thank you~!

PR #158, opened 2021-02-09
Fix column

### LINKED ISSUE #153: Continue completion if the item's word has matched.
For example, In typescript-language-server, we can complete in the below position.

```typescript
import { } from '|';
```

And then, we can type `@` to choose the scoped module. But nvim-compe will close the popup menu.
The reason is that determine returns empty if we type `@`.

It is awesome if nvim-compe continues the current completion.

I have no idea for it for now.

### LINKED ISSUE #156: clangd wants to replace . with -> but it results in .->
## Problem

In C you can access the variables of a struct with `->`
clangd is able to complete variables even if instead `.` is typed.
It deletes the `.` and replaces it with `->` on completion.
When using nvim-compe the `.` doesn't get deleted.
Instead the completion results in `.->`

## Info

NVIM v0.5.0-dev+1073-gb894fbcbe
clangd version 11.0.1

## Minimal init.vim

```vim
call plug#begin($XDG_DATA_HOME.'/nvim/plugins')
    Plug 'hrsh7th/nvim-compe'
    Plug 'neovim/nvim-lspconfig'
call plug#end()

let g:compe = {}
let g:compe.source = {}
let g:compe.source.nvim_lsp = v:true
lua require 'lspconfig'.clangd.setup{}
se completeopt=menu,menuone,noselect
```

## Minimal bug.c

```c
struct Foo {
   int foo;
};

int main() {
   struct Foo *foo;
   foo;
}
```

## Reproduce

1. Create `bug.c`
2. Place cursor between `foo` and `;` like `foo|;`
3. Type `.` to trigger completion
4. Select first completion with `<C-n>`

The line with the completion looks now like

```c
foo.->foo;
```

Instead it should look like

```c
foo->foo;
```

hrsh7th:
Fixes #156
Fixes #153

This PR is buggy now. But it would improve some of the cases.

### DIFF
--- a/autoload/vital/_compe/VS/LSP/CompletionItem.vim
+++ b/autoload/vital/_compe/VS/LSP/CompletionItem.vim
@@ -145,6 +145,7 @@ function! s:_get_expansion(args) abort
     \   . strcharpart(l:current_line, l:request_position.character, strchars(l:inserted_text) + l:overflow_after)
     let l:new_text = l:completion_item.textEdit.newText
     if s:_trim_tabstop(l:new_text) !=# l:inserted
+      " The LSP spec says `textEdit range must contain the request position.`
       return {
       \   'overflow_before': max([0, l:overflow_before]),
       \   'overflow_after': max([0, l:overflow_after]),
--- a/autoload/vital/_compe/VS/LSP/CompletionItem.vim
+++ b/autoload/vital/_compe/VS/LSP/CompletionItem.vim
@@ -146,8 +146,8 @@ function! s:_get_expansion(args) abort
     let l:new_text = l:completion_item.textEdit.newText
     if s:_trim_tabstop(l:new_text) !=# l:inserted
       return {
-      \   'overflow_before': l:overflow_before,
-      \   'overflow_after': l:overflow_after,
+      \   'overflow_before': max([0, l:overflow_before]),
+      \   'overflow_after': max([0, l:overflow_after]),
       \   'new_text': l:new_text,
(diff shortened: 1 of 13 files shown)


autoload/vital/_compe/VS/LSP/CompletionItem.vim
```diff
@@ -146,8 +146,8 @@ function! s:_get_expansion(args) abort
     let l:new_text = l:completion_item.textEdit.newText
     if s:_trim_tabstop(l:new_text) !=# l:inserted
       return {
-      \   'overflow_before': l:overflow_before,
-      \   'overflow_after': l:overflow_after,
+      \   'overflow_before': max([0, l:overflow_before]),
+      \   'overflow_after': max([0, l:overflow_after]),
       \   'new_text': l:new_text,
```
hrsh7th, line 151: should check this.

hrsh7th:
Hm... vim-language-server and sumneko_lua has very complex behavior 😂

Hm... I don't know the vim-language-server supports VSCode or not. So probably I will omit supporting it.

hrsh7th:
This PR removed much of the features temporally (it needed by easy to debug).
I must restore those features.

Now, This PR works as VSCode like, I think.

autoload/vital/_compe/VS/LSP/CompletionItem.vim
hrsh7th: It's ok.

hrsh7th:
Apologize first. Sorry if it broke.

PR #170, opened 2021-02-11
Add omni completion source

cbarrete:
This is a basic source that works, but I think that it should be filetype-specific, because it is only relevant in a few cases (e.g. when editing LaTeX files when https://github.com/lervag/vimtex is installed). However, there isn't a good way for users to configure filetypes for sources right now, since they don't have access to the source metadata.

It would be easy to add a global option and read it in `get_metadata` but that's quite ugly in my opinion.

Ideally, I'd like to configure filetypes via files in an `ftplugin` directory (which only works for Vimscript, [Lua support is planned for 0.6](https://github.com/neovim/neovim/issues/12670), which is still far away) or autocommands. I feel that it shouldn't be the responsibility of a plugin to handle filetype management since (Neo)Vim already has a very good way to do that.
I can open an issue to discuss that if you don't have a plan for it yet.

### DIFF
--- a/lua/compe_omni/init.lua
+++ b/lua/compe_omni/init.lua
@@ -1,0 +1,37 @@
+local compe = require("compe")
+local Source = {}
+
+function Source.new()
+  return setmetatable({}, { __index = Source })
+end
+
+function Source.get_metadata(_)
+  return {
+    priority = 100;
+    dup = 0;
+    menu = '[Omni]';
+  }
+end
+
+function Source.determine(_, context)
+  local fn = vim.bo.omnifunc
+  if fn == '' then
+    return nil
+  end
+  local completion_start_column = vim.api.nvim_call_function(fn, { 1, '' })
+  if completion_start_column == -2 or completion_start_column == -3 then
+    return nil
+  elseif completion_start_column < 0 then
+    completion_start_column = vim.api.nvim_win_get_cursor(0)[2]
+  end
+  return {
+    keyword_pattern_offset = completion_start_column + 1,
+  }
+end
+
+function Source.complete(_, context)
+  context.callback({ items = vim.api.nvim_call_function(vim.bo.omnifunc, { 0, context.input }) })
+end
+
+return Source.new()
+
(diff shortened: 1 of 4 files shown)


hrsh7th:
Thank you very much! I really appreciate contribution!

I'm reviewing.

lua/compe_omni/init.lua
```diff
@@ -0,0 +1,27 @@
...
+  return {
+    priority = 100;
+    dup = 0;
+    menu = '[Omni]';
+  }
+end
+
+function Source.determine(_, context)
+  return compe.helper.determine(context)
+end
```
hrsh7th, line 18: I think we should check `vim.bo.omnifunc ~= ''` and use `vim.api.nvim_call_function(omnifunc, { 1, '' })` to determine `keyword_pattern_offset` instead of `compe.helper.determine`.

lua/compe_omni/init.lua
```diff
@@ -0,0 +1,27 @@
...
+
+function Source.complete(_, context)
+  local fn = vim.bo.omnifunc
+  vim.api.nvim_call_function(fn, { 1, '' })
```
hrsh7th, line 22: The omnifunc design is...

1. First, call omnifunc with `findstart==1` that to determine `pum position`. this is the same with `Source.determine` so should use it on `Source.determine`.
2. Invoke actual completion (it should use on `Source.complete`)

lua/compe_omni/init.lua
hrsh7th: The `Source.determine` function can be return empty dict (it means cancelling completion).
If `omnifunc` is empty, we should return empty on `Source.determine`.

cbarrete:
Thanks for the review, I was hoping for a discussion about filetypes before going into how the omnifunc works, but I'll just open a new issue.
I think we can merge it for all filetypes as it is now (unless you have more comments of course).

hrsh7th:
@cbarrete 
Sorry. I was missing that point.

In compe, the filetype metadata can be confifured by `g:compe.source.omni = { 'filetypes': ['tex'] }`

hrsh7th:
Looks good. I will merge it.

cbarrete:
> In compe, the filetype metadata can be confifured by `g:compe.source.omni = { 'filetypes': ['tex'] }`

Ah thanks, I didn't realize that, I'll still open an issue as I don't think that it's an optimal solution, but we can discuss this :)

PR #177, opened 2021-02-13
Don't show resolve is detail is ''.

ckipp01:
Some servers will respond with the detail field, but with an empty
string. If this happens an empty window is shown. This just adds another
simple check to ensure that the window isn't shown if the detail is ''.

Before:

Screenshot 2021-02-13 at 12 55 51

After:

Screenshot 2021-02-13 at 12 55 22


hrsh7th:
Thank you!

PR #186, opened 2021-02-16
Improve context handling

hrsh7th:
toooooo difficult

### DIFF
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -117,38 +117,26 @@ end
 
 --- complete
 Completion.complete = function(option)
-  if Completion:_should_ignore() then
+  local context = Completion._new_context(option)
+  if context:should_ignore() then
     Async.throttle('display:filter', 0, function() end)
     return
   end
 
-  -- Check the new context should be completed.
-  local context = Completion._new_context(option)
 
   -- Restore pum (sometimes vim closes pum automatically).
-  local is_completing = Completion._is_completing(context)
-  local is_manual_completing = is_completing and not Config.get().autocomplete
+  local is_manual_completing = context.is_completing and not Config.get().autocomplete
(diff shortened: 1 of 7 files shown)


tami5:
Oh yah I've been bothered by seeing lsp suggestions in comments or have snippet suggestion while doing rust magic completion.

If you need any help or have a question just ping me 🙏

hrsh7th:
It can't solve in this PR, I think.

I will check rust-analyzer's behavior.

hrsh7th:
snippet suggestion can be solved (but that is not depends on this PR. We should implement in another PR).

weilbith:
Could you maybe add some docs for this new feature? :upside_down_face:

hrsh7th:
This PR is just a general improvement of context handling.

But one feature was added that will invoke new completion after confirmation by pressing `<CR>`, if if a trigger character has found.

hrsh7th:
The new feature will be useful if you are writing css.

For example, old compe does not invoke new completion if `font-si|<CR>` -> `font-size: |;`.
But new compe will invoke completion so you can look `0em/0px/0em` as completion candidates.

cloggier:
This doesn't seem to work for me. Then I type:
```
fo-si
```
I get `font-size` property completion item, but no snippet. Selecting property completion does not insert `;` at the end of the line nor it triggers another completion for units.

hrsh7th:
Ah. Sorry.
If you use nvim-lspconfig, your css-ls is old.

I'm using `vscode-css-language-server`.
You can install it via `npm i -g vscode-langservers-extracted`

cloggier:
It works, beautiful, thanks a lot!

cloggier:
Ah, is it worth also replacing html and json server from those suggested in lspconfig? Are your versions superior?

hrsh7th:
Yes. All my versions are better than old one.

PR #194, opened 2021-02-19
Implement emoji autocompletion

WhyNotHugo:
This PR adds emoji autocompletion.

I expect this won't be merged in its current state, but rather, expect feedback on it.

I wrote a small script that generates `lua/compe_emoji/data.lua` with all the emoji data -- rather than parse data file in runtime, this offloads the work and has a pre-generated file with everything ready to use.

I've been using this for a day and seems to work fine.

I left the script I used to generate `data.lua` in-repo, since it might be useful for updating the file (e.g.: when new emojis come out). I _wish_ I was proficient enough in `lua` to as to write it in `lua`, but that seems to be beyond me [for now].


elianiva:
I think we should exclude the data from the commit? I think it's better because if someone doesn't want to activate the emoji completion at all, they won't get this massive file. (if it's only for showcase purpose, then ignore me :laughing:)

Also, I think I could help you with the lua script. Gimme a few minutes.

WhyNotHugo:
> I think we should exclude the data from the commit?

The file is generated using a much bigger JSON dump as a source. If it's not included, it needs to be downloaded at some point

Not sure how that would be handled.

The file is big, but just 188kb. I _could_ try to lazy-load in first completion though, so that it's only loaded into memory if it's been activated by the user (I think it's always loaded right now).

sblumentritt:
Why force it into the completion plugin? I always hate plugins which bundle unnecessary things like sources/utilities when there is a clear interface for external plugins. As far as I can see the 'emoji' source does not really need any special plugin support and therefore the source should be a separate plugin in my opinion.

Going the external source approach as multiple benefits:
- The completion plugin repository stays small
- User do not get unnecessary sources / data which they do not want
- The completion plugin is not responsible to maintain the source

There is a reason why the plugin provides an interface to write external sources.

----

Don't take my words to personal 😄

elianiva:
I agree with it being an external extension from this plugin, it'd be easier to maintain and user who doesn't use this feature won't get the data file. But all of that is up to hrsh7th, I'm just saying my opinion :)

hrsh7th:
Thank you very much for your contribution!

Sorry for the my late reply.

I don't want to depend external resources or external build process.

If those dependencies was removed, I can merge it.

WhyNotHugo:
The python script was only used to _generate_ the `data.lua` file. There's not need for it to be in-repo, I just left it since it'll be useful to update emojis when more [inevitably] come out.

I'm not proficient enough in lua to re-write the python script into lua though.

I can either remove the script from the PRO, or just make it a separate plugin, if that sounds best.

hrsh7th:
Hm... It's not hit to startup time.

OK. I can include it but I think the source can be improved.

I will review it.

hrsh7th:
If I have time, I will convert python to Lua (maybe)

lua/compe_emoji/init.lua
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    priority = 80,
+    dup = 1,
+    menu = '[Emoji]'
+  }
+end
+
+Source.determine = function(_, context)
+  return compe.helper.determine(context, {
+    keyword_pattern = [[:\w*$]]
```
hrsh7th, line 15: I think the source should be triggerd only when the line is `\%(\s\|^\)\zs:\w*`.

lua/compe_emoji/init.lua
```diff
@@ -0,0 +1,29 @@
...
+  })
+end
+
+-- Load emoji data from auto-generated file:
+Source._items = require('compe_emoji.data')
```
hrsh7th, line 20: It's better to delay until the first completion.

hrsh7th:
At least, this source can be used as having no external dependencies for the users.

elianiva:
> If I have time, I will convert python to Lua (maybe)

I don't think that's possible if we use lua5.1 or luajit, unicode escape sequence is introduced in lua5.3
I have this [half working script](https://paste.sh/asMHkOHv#wHizUnxuxFuwoEDezTRCgpip) that I stopped working on because I realised it's not possible if I don't use lua5.3

hrsh7th:
Awesome. Thank you.

I'm looking into it.

hrsh7th:
```
local M = {}

M._fetch = function(path)
  return vim.fn.json_decode(vim.fn.readfile(path))
end

M._write = function(path, data)
  local h = io.open(path, 'w')
  io.write(h, data)
  io.close(h)
end

M.to_string = function(chars)
  local fmt = ''
  for _, char in ipairs(chars) do
    fmt = fmt .. ([[\u%s]]):format(char)
  end
  return vim.fn.eval(([["%s"]]):format(fmt))
end

M.update = function()
  local count = 0
  for _, emoji in ipairs(M._fetch('emoji.json')) do
    count = count + 1
    if count > 10 then
      break
    end

    local char = M.to_string(vim.split(emoji.unified, '-'))
    print(char)
  end
end

M.update()
```
It's works.

elianiva:
oh yeah, you could use viml for handling the unicode escape sequence, forgot about that :laughing:

lua/compe_emoji/init.lua
hrsh7th: For example, the PHP/Rust has `Identifier::CONST` syntax.

lua/compe_emoji/gen-data.py
```diff
@@ -0,0 +1,32 @@
+# Generates the emoji data LUA file using this as a source:
+# https://github.com/iamcal/emoji-data/blob/master/emoji.json.
+import json
+
+
+def write_item(char, name, output):
+    output.write("table.insert(items, {")
```
runiq, line 7: Why not construct and return the table directly instead? E.g. in `write_item`:

```python
output.write(f"  {{word = '{char}', abbr = '{char} :{name}:', filter_text = ':{name}:'}},\n")
```
In `write_data`:

```python
output.write("return {\n")
# for loop with write_item
output.write("}\n")
```

This way, you don't do `table.insert` for each and every emoji.

WhyNotHugo: That's actually quite smart. For some reason I though it might be problematic to have such a huge statement, but it's fine. Thanks!

hrsh7th:
LGTM.

I will merge it after solve conflicts.

hrsh7th:
Thank you so much!

WhyNotHugo:
Thanks!

Issue #209, opened 2021-02-23
Show basic type/signature info in the completion menu


YorickPeterse:
When compe presents the user with a list of entries to complete, this list only includes some basic information: the text, the source, and the type of thing (e.g. "Variable"). Other completion engines typically also show a limited form of the signature. 

For example, this is what compe suggests:

![Screenshot from 2021-02-23 18-23-59](https://user-images.githubusercontent.com/86065/108882543-cb701880-7604-11eb-8ee8-1030386eb78b.png)

![Screenshot from 2021-02-23 18-30-19](https://user-images.githubusercontent.com/86065/108882988-33befa00-7605-11eb-873c-c3395d70ace8.png)

For comparison, this is what coc suggests:

![Screenshot from 2021-02-23 18-29-36](https://user-images.githubusercontent.com/86065/108883013-3b7e9e80-7605-11eb-9d92-b1e857a27430.png)

![Screenshot from 2021-02-23 18-29-57](https://user-images.githubusercontent.com/86065/108883015-3b7e9e80-7605-11eb-816c-05da44fa584a.png)

Is this already supported and do I somehow need to enable this, or is this something that could be added to compe? Thanks! :smiley:

hrsh7th:
@YorickPeterse Could you explain how to realize it?
I can't found the `Vec<T, A>` like string in LSP's completion results.

YorickPeterse:
@hrsh7th I think it's the LSP "detail" field. This is how NeoVim's omnifunc does it: https://github.com/neovim/neovim/blob/9d5f842807c56d6265a5557cbb4bf3e951210425/runtime/lua/vim/lsp/util.lua#L438

The field is described in https://microsoft.github.io/language-server-protocol/specification#textDocument_completion under `
export interface CompletionItem` (sadly can't link to it directly):

```
	/**
	 * A human-readable string with additional information
	 * about this item, like type or symbol information.
	 */
	detail?: string;
```

hrsh7th:
I don't think it is `detail` field.
The rust-analyzer does not returns `Vec<T, A>` as detail field, I think.

hrsh7th:
The coc shows `detail` field as floating documentation's first-line in default.

hrsh7th:
Sorry. If I had updated rust-analyzer, it was returned.

Hm... I will consider it.

tami5:
@hrsh7th this is so useful and saves me from `<cr><esc>K` see the function signature then `i` :laughing:

[...]

PR #219, opened 2021-02-25
RFC: Temporary detail field solution before  LSP 3.17

hrsh7th:
Use the detail field instead of the menu field if manual completion is triggered.

Related #209

### DIFF
--- a/lua/compe/completion.lua
+++ b/lua/compe/completion.lua
@@ -195,7 +195,11 @@ Completion._display = guard(function(context)
           item.word = item.original_word
           item.abbr = item.original_abbr
           item.kind = item.original_kind or ''
-          item.menu = item.original_menu or ''
+          if context.manual and item.detail then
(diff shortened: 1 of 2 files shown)


hrsh7th:
I think this solution is temporary because LSP3.17 has a `CompletionItemLabelDetails` we should use it.

hrsh7th:
@YorickPeterse What do you think about it?

YorickPeterse:
@hrsh7th This looks good. However, I can see users perhaps being confused by the difference in information shown between manual and auto completions. As you mentioned in other issues, it's also an extra option to deal with. 

I've personally been experimenting with my own completion solution, mostly so I don't have to bother others with my use case. So if nobody else needs this information, perhaps you can simply leave it out :smiley:

hrsh7th:
Thank you for your comment.
OK. I will continue to consider it.

PR #223, opened 2021-02-26
Add kind mapping

ray-x:
This is a port from complete-nvim feature which will show LSP kinds to a nerdfont
e.g {["Function"] = ""; ["Interface"] = "擄";}

Screen Shot 2021-02-26 at 4 14 59 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ let g:compe.max_abbr_width = 100
 let g:compe.max_kind_width = 100
 let g:compe.max_menu_width = 100
 let g:compe.documentation = v:true
+let g:compe.kind_mapping =  {"Function" = ""; "Interface" = "擄";};
+
 
 let g:compe.source = {}
 let g:compe.source.path = v:true
@@ -127,6 +129,7 @@ require'compe'.setup {
   max_kind_width = 100;
   max_menu_width = 100;
   documentation = true;
+  kind_mapping = {["Function"] = ""; ["Interface"] = "擄";};
 
   source = {
     path = true;
@@ -207,7 +210,6 @@ inoremap <silent><expr> <C-d>     compe#scroll({ 'delta': -4 })
 - [zsh](https://github.com/tamago324/compe-zsh)
 - [conjure](https://github.com/tami5/compe-conjure)
 - [dadbod](https://github.com/kristijanhusak/vim-dadbod-completion)
-
 
 ## FAQ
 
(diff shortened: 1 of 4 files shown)


hrsh7th:
Thank you for your contribution.

I think this feature is unnecessary because we already have https://github.com/onsails/lspkind-nvim .

I'm considering it to support by nvim-compe itself. But it still needs more time.

ray-x:
In fact,  https://github.com/onsails/lspkind-nvim  is work out of box. As it replace language server by 
```lua
require('vim.lsp.protocol').CompletionItemKind = symbols
```

But The way Lspkind-nvim update kind is by attach the nerd font in front of kind. So the kind will be 

```
{ " Text", "ƒ Method", " Function", ....}
```

But what I want is 

```
{ "", "ƒ", " ", }
```
Which is similar to vscode.
Check the behavior below:
iskind:
Screen Shot 2021-02-26 at 5 10 18 pm


vscode:

Screen Shot 2021-02-26 at 5 09 54 pm

hrsh7th:
I'm now using lspkind with following config and it worked as you expected.

```

if dein#tap('lspkind-nvim')
  lua require'lspkind'.init({ with_text = false })
endif
```

ray-x:
Well, 2~3 lines of change in compe or introduce a plugin for it...
I would probably change my init.lua for this if the PR rejected.

hrsh7th:
To be honest, I think this feature needs more consideration.

For example, Currently, the buffer source results don't have kind. But user will want this configuratuon affect to buffer source, I think.

ray-x:
As I only need a hard code kind icon. So I can easily do this in init.vim while I waiting for this feature :
require('vim.lsp.protocol').CompletionItemKind = {'', 'ƒ', '', '', 'ﴲ', '', '', 'ﰮ', '', '', '', '', '了', '', '﬌', '', '', '', '', '', '', '', 'ﳅ', '', '', ''}

What would be ideal that:
1) Type icon in the font (same as vscode, would be great with color) and then the item.abbr
Screen Shot 2021-02-26 at 7 22 16 pm

2) Set a default for those does not have kind.

hrsh7th:
I think the ideal solution like as the following.

```
local compe = require'compe'

local source = {}

source.complete = function(_, context)
  local items = {}
  table.insert(items, {
    word = 'something';
    kind = compe.helper.kind.Keyword;
  })
  context.complete({
    items = items
  })
end
```

It means we should define `kind table` in nvim-compe built-in.
Currently, it depends on the LSP but it should be able to use in some other sources.

PR #236, opened 2021-03-04
Enable using snippets.nvim for LSP snippets

runiq:


hrsh7th:
Looks good. Thank you!

runiq:
Augh dammit, I forgot to update the README and docs. Give me a second for a followup PR…

sharksforarms:
@runiq Works great! Thanks!

p00f:
image
I get mismatched indentation when I use snippets.nvim with clangd (sumneko isn't suggesting snippets for some reason and I don't use any other snippet-providing server)

runiq:
@p00f Dumb question: How do I invoke that snippet…?

p00f:
type `if` somewhere in a function?

hrsh7th:
The clangd's `if` snippet becomes the following results in vsnip.

```
  if<CR>
  ↓↓↓
  if (condition) {
  statements
  }
```

runiq:
Ah, I see the issue now.

Both sumneko and rust-analyzer "pre-indent" the textEdits they get from the langserver, clangd does not. There's an implicit assumption here:

1. Sumneko and rust-analyzer expect the editor to apply textEdits as-is (i.e. without reindenting).
2. clangd expects the editor to reindent whatever it receives from the langserver.

This has been clarified with the [`InsertTextMode` setting in LSP 3.16](https://microsoft.github.io/language-server-protocol/specification#textDocument_completion). I'm not sure how difficult it would be to support that, though.

hrsh7th:
I checked the snippets.nvim's behavior. That's works the same as vsnip.

It just clangd's problem, I think.

p00f:
but indentation is correct when i use vsnip

hrsh7th:
The CompletionItem that returned from clangd is the following.

```
'if (${1:expression}) {${0:statements}}'
```

`${0:statements}` haven't been indented.

Yes. Your screenshot is the wrong result but I can't repro it with snippets.nvim.
I don't familiar with snippts.nvim. Sorry...

@runiq Could you check it?

hrsh7th:
Expected (vsnip's result):
https://github.com/hrsh7th/nvim-compe/pull/236#issuecomment-792701779

Actual (@p00f 's result):
https://github.com/hrsh7th/nvim-compe/pull/236#issuecomment-791967630

runiq:
@hrsh7th On it!

This is the completionItem I get from the langserver:

```
{ 
  filterText = "if", 
  insertText = "if (${1:expression}) {\n${0:statements}\n}", 
  insertTextFormat = 2, 
  kind = 15, 
  label = " if (expression) {statements}", 
  score = 4, 
  sortText = "3f800000if", 
  textEdit = { 
    newText = "if (${1:expression}) {\n${0:statements}\n}", 
    range = { 
      end = { 
        character = 2, 
        line = 1 
      }, 
      start = { 
        character = 1, 
        line = 1 
      } 
    } 
  } 
}
```

And this happens when I apply that item with compe and snippets.nvim:

https://user-images.githubusercontent.com/620981/110323704-4417a180-8015-11eb-9092-8aa5f3916ad5.mp4

That's not a snippets.nvim problem, I think—doesn't vsnip do the same?

@p00f I think at this point I need to see your clangd and snippets.nvim setup, sorry :(

p00f:
okay, thanks for looking into this!

hrsh7th:
https://github.com/hrsh7th/nvim-compe/pull/236#issuecomment-792737621
I had the same result with snippets.nvim

runiq:
@p00f Do you maybe have another plugin that messes with indenting?

runiq:
Oooh, could it be that clangd reindents things using a [`document/onTypeFormatting` request](https://microsoft.github.io/language-server-protocol/specification#textDocument_onTypeFormatting)? Because the nvim LSP client currently doesn't support that, while the clangd server does.

hrsh7th:
https://gitter.im/neovim/neovim?at=604626cde562cf54aca8283b

@p00f seems to not have any snippets.nvim config.

So the problem caused by `s:simple_expand_snippet`.

I can improve it but I think we shouldn't use nvim-compe without any snippet engine.

PR #239, opened 2021-03-05
Add padding to documentation window

### LINKED ISSUE #169: Make the gap between completion menu and floating window removable
When navigating through completion items, there's a secondary floating window showing definitions (at least for LSP items). There's a small gap between the completion menu and that window, and sometimes there's a character from the background buffer in that gap, making it a little harder to read the floating window sometimes.

I imagine it's someone's preference to have it that way, so it would be cool if that gap was removable via configuration.

cloggier:
Closes #169

### DIFF
--- a/autoload/compe/documentation.vim
+++ b/autoload/compe/documentation.vim
@@ -45,7 +45,11 @@ function! compe#documentation#open(document) abort
 
     " Ensure normalized document
     if !has_key(s:cache, l:state.document)
-      let s:cache[l:state.document] = split(s:MarkupContent.normalize(l:state.document), "\n")
+      let l:padded_document = []
+      for line in split(s:MarkupContent.normalize(l:state.document), "\n")
+        call add(l:padded_document, ' ' . line . ' ')
+      endfor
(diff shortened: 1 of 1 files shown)


cloggier:
@hrsh7th Very naive way to add padding. Does not work fully, won't insert spaces on lines starting with 

    ```
for some reason. Any idea why? Do you even like that way? Or should it be solved with creating extra window for padding only?

hrsh7th:
Thank you for your contribution.

Yes. I was tried this approach too but I met the same problem.

I think it would be occurred by https://github.com/hrsh7th/vim-vital-vs/blob/master/autoload/vital/__vital__/VS/Vim/Syntax/Markdown.vim#L23 .
I also tried to improve the fenced language block regex but I couldn't...

cloggier:
Another option is to create extra window for padding. Would such solution be acceptable for you?

hrsh7th:
Yes. I think another window is better solution (due to `wrap` option)

But that approach

cloggier:
Superseded by https://github.com/hrsh7th/nvim-compe/pull/249

PR #240, opened 2021-03-07
Add padding

hrsh7th:
Add padding for the documentation window.

Unfortunately, It will cause a flicker...

### DIFF
--- a/autoload/compe/documentation.vim
+++ b/autoload/compe/documentation.vim
@@ -14,6 +14,15 @@ call setbufvar(s:window.get_bufnr(), '&bufhidden', 'hide')
 call setbufvar(s:window.get_bufnr(), '&buflisted', 0)
 call setbufvar(s:window.get_bufnr(), '&swapfile', 0)
 
+let s:padding = s:FloatingWindow.new()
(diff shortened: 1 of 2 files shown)


autoload/compe/documentation.vim
```diff
@@ -55,24 +69,31 @@ function! compe#documentation#open(document) abort
...
-      \   'row': l:pos[0] + 1,
-      \   'col': l:pos[1] + 1,
-      \   'width': l:size.width,
-      \   'height': l:size.height,
-      \ })
-      silent call s:Window.do(s:window.get_winid(), { -> s:Markdown.apply() })
-    endif
+    call s:window.close()
+    call s:padding.close()
+    silent call s:window.open({
```
hrsh7th, line 84: In my environment, neovim will be leaving the rendering garbage so I added `s:window.close` and `s:padding.close` explicitly in here.
Then the documentation window will be display correctly.

However, this approach occurs a flicker, unfortunately.

hrsh7th:
This PR needs https://github.com/neovim/neovim/pull/14020

cloggier:
Could that be of help: https://github.com/neovim/neovim/pull/12444?

PR #249, opened 2021-03-11
Add padding and bunch of improvements

hrsh7th:
1. Support floating documentation padding
- It's difficult but I solved it by https://github.com/hrsh7th/vim-vital-vs/commit/101e1819bd641ba29cadfbe353eae0d678cce8c5

2. Allow prefix unmatched item

3. If multiple sources triggered by character, use all sources.

4. Helper improvements (maybe)
- Sorry... if it's broken.


hrsh7th:
@cloggier Could you test it? I think it is really improved.

cloggier:
Works very nice after quick look! Small issue with displaying emojis?

Screenshot 2021-03-11 at 13 25 40

hrsh7th:
@cloggier You should set correct value for `set ambwidth`. If your terminal set upped to `Use double width for multibyte chars`, You should set `set ambwidth=double`.

hrsh7th:
I will merge it in few days of testing.

cloggier:
> set ambwidth=double

Doesn't seem to be helping at all (`ambiwidth=double`). Might be my terminal. I don't think I care enough, but will look into it later.

Also, found another example of markdown escape character being displayed:
Screenshot 2021-03-11 at 14 33 00

hrsh7th:
@cloggier The `set ambwidth` option depends on your terminal setting. Please try `double` or `single` both.

And escape char problem already fixed now.

hrsh7th:
wait. Your screenshot had `gap` between completion-menu and documentation window so I guess your branch is master.

hrsh7th:
Some bugs that relate to rendering markdown are fixed now.

IMO, Now compe has no markdown rendering problem. I will check the basic completion behaviors before merge it.

PR #256, opened 2021-03-14
add preselect confirm mode

oberblastmeister:
I added a new confirm function in lua because I can't access the `Config` in vimscript. Also lua is much easier to write and probably faster as a side benefit.

### DIFF
--- a/lua/compe/init.lua
+++ b/lua/compe/init.lua
@@ -1,5 +1,6 @@
 local Debug = require'compe.utils.debug'
 local Callback = require'compe.utils.callback'
+local api = vim.api
 local Completion = require'compe.completion'
 local Source = require'compe.source'
 local Config = require'compe.config'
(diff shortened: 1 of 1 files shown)


hrsh7th:
I think we should implement `compe#confirm({ 'fallback': '<Tab>', 'select': v:true })` instead of this solution.
I will implement it.

This PR has no vim side implementation so I can't merge it. Sorry.

hrsh7th:
@oberblastmeister Could you test select-option branch?

oberblastmeister:
just curious, why must it be a vimscript implementation?

hrsh7th:
Because the Lua's keymapping is problematic, I think.

cloggier:
> Because the Lua's keymapping is problematic, I think.

In what way? As in, too verbose?
Would be nice not to have any `viml` here. :)

hrsh7th:
For example, nvim-compe is providing the `compe#complete` that is expected for use in `<expr>` mapping.
I think the `<expr>` mapping can't call directly in Lua so after all, it is useful if it is written in Vim script.

cloggier:
I think it does work, one can do that:

```lua
vim.api.nvim_set_keymap("i", "<CR>", "v:lua.OnEnter()", {noremap = true, expr = true})
```

which is what I am doing. Would be possible to just call `compe.Complete` or something like that, I believe.

hrsh7th:
Hm... I can't think it is better... (It has no difference for my eyes)

## Now
```lua
vim.api.nvim_set_keymap("i", "<CR>", "compe#complete('<CR>')", {noremap = true, expr = true})
```

## Proposed
```lua
vim.api.nvim_set_keymap("i", "<CR>", "v:lua.compe.complete('<CR>')", {noremap = true, expr = true})
```

cloggier:
From this perspective it really changes very little, but would allow for getting rid of viml completely.

oberblastmeister:
I think both implementations are okay on the outside and expr mappings definately work with lua. However, lua is trying to be a first class citizen in neovim and since this plugin doesn't support `vim` I feel like it makes sense to use lua and get rid of viml. In addition, lua keymaps and autocommands are coming so it can make the above snippet even nicer, as we can directly map to lua functions. To me, lua is objectively a better language and we are using `nvim` so we should use it. Those were just my (slightly opinionated) thoughts. Anyways, thanks for your work on this plugin because it has made the completion experience very nice and smooth.

PR #258, opened 2021-03-15
Add `compe#confirm({ 'select': v:true, 'keys': '<CR>' })`

### LINKED ISSUE #254: disable default `<CR>` mapping
By default when you click `<CR>` an item from the completion menu will be selected. Is there any way to disable it?

hrsh7th:
Fixes #254 
instead of #256


oberblastmeister:
This works perfectly. Thanks

Issue #260, opened 2021-03-16
Wrong markdown rendering when showing method documentation


baco:
When displaying documentation of a method, the signature gets interpreted as Markdown instead of being treated as a code snippet. As seen in the first screenshot `foo(*args, **kwargs)` is treated as if the intention was to make bold text using the `*` character around words:
![Screenshot from 2021-03-16 12-10-44](https://user-images.githubusercontent.com/102958/111333376-63d04a80-8651-11eb-872a-a37d91cf3f55.png)
_(screenshot using nvim-compe + python-language-server)_

The **correct display** of that method's documentation would be something like this:
![Screenshot from 2021-03-16 12-12-17](https://user-images.githubusercontent.com/102958/111333396-69c62b80-8651-11eb-8ac3-3debe509208e.png)
_(screenshot taken with CoC + jedi-language-server, for comparison)_

Also, the docstring of the method got stripped away.

The Language Server used for first Python example screenshot is `pyls`, the Language Server used in **the correct** screenshot is `jedi-language-server`.

**Update:**
Changing the Language Server, for another provider for Python, doesn't solve the issue... if not, makes it somehow worse because Markdown quotation appears on the info window now:
![Screenshot from 2021-03-16 17-35-14](https://user-images.githubusercontent.com/102958/111376386-2635e680-867e-11eb-8e57-a10c7cc6554e.png)
_(screenshot using nvim-compe + jedi-language-server)_

The Language Server used in this new screenshot is again `jedi-language-server`.

hrsh7th:
I've updated markdown rendering. https://github.com/hrsh7th/nvim-compe/commit/8c0b843faba2a698836bee42ac793943dee91192

I have no python environment. Sorry.
Could you test with latest master?

baco:
ok, trying out:

```shell
baco@terrapin ~/s/d/n/p/m/s/nvim-compe ((ab623db5…))> git lg HEAD~3...
* ab623db (HEAD, origin/master, origin/HEAD) Improe perf — hrsh7th (2 hours ago)
* 8c0b843 Improve markdown rendering — hrsh7th (3 hours ago)
* b5decc5 Improve a bit — hrsh7th (4 hours ago)
```

It renders more information now, like it is not truncated the hole docstring of the python method, but I still get a misinterpretation of code to render:

![image](https://user-images.githubusercontent.com/102958/111621765-2261ab80-87c7-11eb-9e4d-dacafd28d99d.png)
[...]

PR #270, opened 2021-03-18
Enclose item documentation as Markdown code-block

baco:
I think this change may solve #260 

Screenshot from 2021-03-18 16-27-55

Star (`*`) characters don't get interpreted as Markdown bold syntax and Python's built-in types get syntax color as if interpreted as code.


hrsh7th:
Sorry. It is not the correct solution.

I'm tackling it.

folke:
This should be fixed as part of #426

PR #285, opened 2021-03-23
multiple filetypes for Snippets.nvim Fix #283

### LINKED ISSUE #283: Recognise multible filetypes for snippets_nvim
I use the plugin Snippets.nvim for sinppets. If i have a buffer with one filetype all is fine, but if i have multible filetypes it doesn't find the corresponding snipptes.

I looked in `lua/compe_snippets_nvim/init.lua` that the filetype is just used as a string:

```lua
local snippets_list = vim.tbl_extend ('force',
  snippets_nvim.snippets._global or {},
  snippets_nvim.snippets[vim.bo.filetype] or {}
)
```

but the filetypes are speparted by a `.` in `vim.bo.filetype`. In `:h ft` you can find:

```
When a dot appears in the value then this separates two filetype
names.  Example:
	/* vim: set filetype=c.doxygen : */ ~
This will use the "c" filetype first, then the "doxygen" filetype.
This works both for filetype plugins and for syntax files.  More than
one dot may appear.
```
I changed the code to:

```lua
local snippets_list = snippets_nvim.snippets._global or {}
local filetypes = vim.split(vim.bo.filetype, ".", true)
for _ , ft in ipairs(filetypes) do
  snippets_list = vim.tbl_extend ('force',
    snippets_list,
    snippets_nvim.snippets[ft] or {}
  )
end
```

and now it is working.

I saw that `vim.bo.filetype` is used "wrong" in the same way in other locations.

z4m1n0:
Fixed wrong use of `vim.bo.filetype` in `lua/compe_snippets_nvim/init.lua`.

### DIFF
--- a/lua/compe_snippets_nvim/init.lua
+++ b/lua/compe_snippets_nvim/init.lua
@@ -25,10 +25,15 @@ end
 function Source.complete(_, context)
   local items = {}
 
-  local snippets_list = vim.tbl_extend ('force',
-    snippets_nvim.snippets._global or {},
-    snippets_nvim.snippets[vim.bo.filetype] or {}
-  )
+  local snippets_list = {}
+  local filetypes = vim.split(vim.bo.filetype, ".", true)
+  table.insert(filetypes, '_global')
+  for _ , ft in ipairs(filetypes) do
+    snippets_list = vim.tbl_extend ('force',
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you~.

PR #297, opened 2021-03-29
WIP: Replace feature

hrsh7th:
Hm... Probably, we should introduce `replace` option.

`let g:compe.replace = v:true`

If this enabled, the after line should be removed.

We should consider it more deeply.

### DIFF
--- a/autoload/compe.vim
+++ b/autoload/compe.vim
@@ -41,10 +41,12 @@ endfunction
 "
 " confirm
 "
+let g:___compe_confirm_option = {}
+inoremap <silent><nowait> <Plug>(compe-confirm) <C-r>=luaeval('require"compe"._confirm_pre()')<CR><C-y><C-r>=luaeval('require"compe"._confirm(_A)', g:___compe_confirm_option)<CR>
 function! compe#confirm(...) abort
-  let l:completeopts = split(&completeopt, ',')
+  " Check completeopt
   for l:opt in ['menuone', 'noselect']
-    if index(l:completeopts, l:opt) == -1
+    if stridx(&completeopt, l:opt) == -1
       echohl ErrorMsg
       echomsg '[nvim-compe] You must set `set completeopt=menuone,noselect` in your vimrc.'
       echohl None
@@ -55,7 +57,7 @@ function! compe#confirm(...) abort
   let l:select = get(l:option, 'select', v:false)
   let l:selected = complete_info(['selected']).selected != -1
   if mode()[0] ==# 'i' && pumvisible() && (l:select || l:selected)
-    call luaeval('require"compe"._confirm_pre()')
+    let g:___compe_confirm_option = l:option
     let l:confirm = ''
     let l:confirm .= l:select && !l:selected ? "\<C-n>" : ''
     let l:confirm .= "\<Plug>(compe-confirm)"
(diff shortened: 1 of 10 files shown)


tzachar:
This looks like it is only in effect when the suggested item perfectly matches the current line suffix.
I think a different approach would probably be better. Take a look at https://github.com/tzachar/compe-tabnine/pull/8, which uses the new line suffix returned from TabNine. In this account, TabNine actually does the heavy lifting of supplying the old_suffix and new_suffix which the source can then use to replace the suffix with the new item and the new suffix. However, this only happens when the user presses `<CR>`, and the `Source.complete` function is called. It would be amazing if this could be implemented by compe, in general, for each source supplying a new_suffix and old_suffix for each completion item. 

The current infrastructure you already implemented can surely be leveraged to that end.
WDYT?

tami5:
It would be nice if the replace doesn't cut the text to the main register + or \*

hrsh7th:
@tzachar I'm updating this for following VSCode behavior.

hrsh7th:
> It would be nice if the replace doesn't cut the text to the main register + or *

@tami5 This solution is using `setline` so this doesn't break any registers.

hrsh7th:
This PR supports real-time replacing behavior but probably this is overkill...

tami5:
> This PR supports real-time replacing behavior but probably this is overkill...

What's real-time replacing behavior, sounds very interesting

hrsh7th:
1. `wo|<C-n>rd`
2. `wording|<C-p>`
3. `wo|rd`

It means the line that after cursor will be managed by compe.
But it has a lot of side-effect so I don't know if I may or not may merge it.

tami5:
Oh that looks really nice and advance I like it. Yes, I agree. @hrsh7th I think we should have dev branch and only merge to master after new stuff mature. That way we don't have unexpected behaviors shock users. I'm thinking of doing the same for telescope where you have something like 1.0 branch and have fun experimenting and breaking stuff

hrsh7th:
I've changed the approach.

This PR has now highlight replacement ranges instead of real-time modification.


https://user-images.githubusercontent.com/629908/112940305-ed127180-9167-11eb-9ea0-0cf2028b8963.mp4

And we should provide `compe#confirm({ 'replace': v:true })` that replaces highlighted ranges.

This approach similar to what is VSCode does.

```
inoremap <silent><expr> <CR> compe#confirm({ 'replace': v:false })
inoremap <silent><expr> <C-l> compe#confirm({ 'replace': v:true })
```

hrsh7th:
Implemented this for temporary solution.

We should remove `Source.confirm` and add VSCode expansion implementation as first class (Currently, `Source.confirm` is source-specific things).

hrsh7th:
Hmm... We should apply a lot of refactoring before implementing this...

tzachar:
Hi @hrsh7th 
Any way to incorporate the old_suffix / new_suffix protocol ? E.g., use the new_suffix to preview the replaced text?

hrsh7th:
I think new prefix should be added to item.word or emulate LSP's textEdit feature.

old_suffix should be used for calculating each item's replace_range, I think.

Anyway, this feature still under development and It definitely change API.

tzachar:
What you describe is exactly how this is implemented in compe-tabnine, thanks to @freehaha:

```lua
function Source.confirm(self, option)                                                   
  local item = option.completed_item                             
  local pos = api.nvim_win_get_cursor(0)                                                
  local row = pos[1] - 1                                                                
  local col = pos[2]                                                                    
  local len = string.len(item.user_data.old_suffix)                                     
  api.nvim_buf_set_text(0, row, col, row, col+len, {item.user_data.new_suffix})                                
end                                                                                     

Adding the new_suffix / old_suffix protocol to all sources, and generating default ones for sources which do not supply them looks like a sane way.

```

PR #298, opened 2021-03-29
Try to fix incomplete handling

### LINKED ISSUE #284: No completion options for specific LSP
I've switched from completion-nvim and now I am not getting completion options with a specific LSP. (It works fine for pyright).

I have been comparing the traces the LSP shows and I think I found the difference.
I noticed in the `textDocument/completion` request nvim-compe sends out, there is a `context.triggerKind` field, which is not present in completion-nvim.

For some unknown reason this seems to prevent the LSP from actually returning any completion options.
Is there a way to remove this field from the completion request or is this a bug in the LSP or LSP config?

hrsh7th:
Fixes #284 (maybe)

@zegervdv Could you test with it?

### DIFF
--- a/lua/compe/helper.lua
+++ b/lua/compe/helper.lua
@@ -140,7 +140,7 @@ Helper.convert_lsp = function(args)
 
   return {
     items = completion_items,
-    incomplete = response.isIncomplete or false,
+    incomplete = response.isIncomplete,
     keyword_pattern_offset = keyword_pattern_offset;
   }
 end
--- a/lua/compe/source.lua
+++ b/lua/compe/source.lua
@@ -163,7 +163,7 @@ Source.trigger = function(self, context, callback)
 
       self.revision = self.revision + 1
       self.status = 'completed'
-      self.incomplete = result.incomplete or false
+      self.incomplete = type(result.incomplete) == 'boolean' and result.incomplete or (self.incomplete and incomplete)
       self.keyword_pattern_offset = result.keyword_pattern_offset or state.keyword_pattern_offset
       self.trigger_character_offset = state.trigger_character_offset
       self.items = self:_normalize_items(context, result.items)
--- a/lua/compe/source.lua
+++ b/lua/compe/source.lua
@@ -156,7 +156,7 @@ Source.trigger = function(self, context, callback)
       end
 
       -- Continue current completion
-      if count > 0 and #result.items == 0 then
+      if (incomplete or count > 0) and #result.items == 0 then
         self.status = 'completed'
         return callback()
       end


zegervdv:
it still only launches the first request.

hrsh7th:
@zegervdv Sorry. I've updated PR. Please re-test it.

zegervdv:
no worries, I really appreciate you putting time in this.

I've tried this version, but still only one request is going out.
I added some debug prints:
```
diff --git a/lua/compe/source.lua b/lua/compe/source.lua
index 9588a48..dfba338 100644
--- a/lua/compe/source.lua
+++ b/lua/compe/source.lua
@@ -164,6 +164,11 @@ Source.trigger = function(self, context, callback)
       self.revision = self.revision + 1
       self.status = 'completed'
       self.incomplete = type(result.incomplete) == 'boolean' and result.incomplete or (self.incomplete and incomplete)
+      if self.incomplete then
+        print("Source ".. self.name .." is incomplete")
+      else
+        print("Source ".. self.name .." is complete")
+      end
       self.keyword_pattern_offset = result.keyword_pattern_offset or state.keyword_pattern_offset
       self.trigger_character_offset = state.trigger_character_offset
       self.items = self:_normalize_items(context, result.items)
```
And I do not see the nvim-lsp source printing either. ( I do get spell and buffer printing something)

hrsh7th:
@zegervdv Could you provide your key sequence and what is your expected?

```
class Main {
  publi static void main(String[] args) {
    |
  }
}
```

1. `S|`, should send completion request
2. `Sy|`, if step-1 received `isIncomplete=true`, should send new completion request
...

zegervdv:
well, if step 1 received `isIncomplete=true` or `Results: null` it sends a new request.

hrsh7th:
Hm... I think if step-1 received `Results: null`, we shouldn't send a new request...?
If we receive `isIncomplete=true`, we should send a request, it's understandable.

OK. I will investigate VSCode implementation... it sounds like a VSCode's bug...

hrsh7th:
I checked the VSCode implementation. After all, I think the server must return `isIncomplete=true` first.
If not, the completion-engine does not handle the source as `isIncomplete=true`.

zegervdv:
But both vscode and nvim-completion send new requests after receiving `Results: null`.

hrsh7th:
`isIncomplete = true`(1) -> `Results: null`(2) is handled as `isIncomplete = true`.
`Results: null`(1) is not handled as `isIncomplete = true`.

If compe receives `isIncomplete=true` first, compe should send request again.

hrsh7th:
Sorry for my poor English... 😢

zegervdv:
ok, so what I see is the first 2 or 3 requests get `Results: null`, and then eventually the server responds with `isIncomplete=true`.
So that means VSCode for some reason keeps sending requests to the server.
But that is not how it should be according to the LSP specification, right?

zegervdv:
Could it be that VSCode sees there are no completion options (yet) and just keeps trying? 
So if `Results: null` you just assume there was not enough info to provide completion yet, so try again later (e.g. after typing some more characters).
So it would be a separate path from the `isIncomplete=true`.

hrsh7th:
I think VSCode seems to just ignore such a situation.
https://github.com/microsoft/vscode/blob/main/src/vs/editor/contrib/suggest/suggest.ts#L214

zegervdv:
but then where do those new requests come from?

hrsh7th:
I don't know at the moment. I want to know it.
Could you provide VSCode extension repo...?

hrsh7th:
I didn't want to know the server but VSCode extension, I want.

Hmm... I can't read the all of VSCode implementation...

I will investigate it more deeply but I don't know if I get the root cause or not, Sorry.

zegervdv:
Yes, but the extension is part of the software, so I cannot share it.
Again, thanks for the support, I'm not making it easy for you :).

When I type `<c-space>` after typing some more characters I do get completions, so would there be a way to re-trigger the completion request after having typed some characters? Basically do that `<c-space>` more automatically?

hrsh7th:
Thank you for words.

The `min_length` option could solve your problem...?

The compe#confirm sould be used for manual completion.

zegervdv:
Even if I set it to 7 chars. it still returns `null` the first time. Maybe this is just a bug in the server which is hidden by either a bug in VScode or by the behaviour of the extension.

hrsh7th:
Probably, I understood it.
It's compe's bug.

I will fix it but it's dfficult so take a time.

zegervdv:
Thanks. Let me know if you need me to test something.

hrsh7th:
@zegervdv Ah. Sorry. I think this problem should be solved in latest master.

PR #314, opened 2021-04-03
Improve LuaSnip completion

carlitux:
Using `name` as word breaks completion displaying a lot of options.

This proposal use name and dscr to improve documentation and keep trigger as text.

Screenshot from 2021-04-03 15-30-00


hrsh7th:
@L3MON4D3 Could you review it?

L3MON4D3:
Looks good, `trigger` is the better choice for word as it is probably unique.

PR #315, opened 2021-04-04
Allow sources to specify per-item priorities

tzachar:
This allows a source to specify the internal order of completion items
originating from this same source

### DIFF
--- a/lua/compe/source.lua
+++ b/lua/compe/source.lua
@@ -361,7 +361,9 @@ Source._normalize_items = function(self, _, items)
     -- internal properties
     item.item_id = self.item_id
     item.source_id = self.id
-    item.priority = metadata.priority or 0
+    if not item.priority then
+       item.priority = metadata.priority or 0
+    end
     item.sort = Boolean.get(metadata.sort, true)
     item.suggest_offset = item.suggest_offset or self.keyword_pattern_offset
 


hrsh7th:
@tzachar I think we don't need this feature.

We can use `sort = false` instead.

tzachar:
Well, this does not solve my issue.
Let me explain; TabNine returns completion items and sometimes attached the prediction strength in each case. In these cases, it makes sense to override the default sorting behaviour of compe, and signal that the returned completion items should not be sorted but presented as a block.
Setting `sort = false` does not help, no matter where I set it. 
The only workaround I found was to bump each item's priority by a fraction, so that compe, when sorting them, will not reorder the returned items.
However, this only works when applying the above patch.

If there is any other way, I'll gladly adopt it.

hrsh7th:
The `sort = false` means `The source should sort items`.
So if tabnine has own sort score, the source should sort items by the source itself.

The `path` source uses the `sory = false`

tzachar:
ok, seems like I misplaced the `sort = false`. 
Looks like it works. I'll update my plugin
10x

tzachar:
Sorry, I was wrong. tested with my patch applied..
Setting `sort = false` inside the `source.get_metadata` does not work.
completion items are reordered by compe regardless of what I do.

tzachar:
forgot to reopen...

tzachar:
Looks like the problem stems from this line: https://github.com/hrsh7th/nvim-compe/blob/f167a1384c47d7eb632eb27e90cdf7dfdb7169ff/lua/compe/matcher.lua#L317

The completion items returned from compe-tabnine always reach this line when compared, as all items have the same priority.
As no one sets the item.index key, items from the same source are randomly sorted.

A different solution would be to actually set the item.index key based on the order of the items as retuend from the source.

hrsh7th:
@tzachar Thank you for your investigation.

It's a bug. I will fix it.

tzachar:
4b3ade1 fixes the issue.
Closing the PR.

PR #325, opened 2021-04-08
Support Environment Variables in Path Completion

### LINKED ISSUE #320: Feature Request: add an option to complete paths from the working directory
when you trigger path completion, with "." or "/", it's either relative to the root folder ( / ) or the current file. 

Most of the time I open nvim at the root of my current project and all paths are defined relative to this. With R projects I will tipically edit a "R/test.R" file and if I want to source another file I'll do

```r
source("R/other.R")
```

it would be great to have the option to have `nvim-compe` complete this path from nvim working directory.

I see several possibilities for it:

- always trigger file completion for path in strings  (I had that with completion-nvim with:

```lua
g.completion_chain_complete_list = {
  default = {
    default = {
      { complete_items = {'lsp', 'snippets'} },
   },
   string = {
      { mode = 'file' },  -- use ins-complete "files" (search ins-compl in completion-nvim help)
    },
  },
}

```

- use a special trigger that would disappear after completion (e.g. "," triggers completion and after selection we get "R/other.R" without the trigger char

### LINKED ISSUE #322: Parent Directory Prefix
Not a big issue, but probably worth mentioning. There's a small bug in path completion for the prefix matching of a parent directory with `../`. The relevant code (`lua/compe_path/init.lua:L69`) has `prefix:match('%../$')` instead of `prefix:match('%.%./$')`, so it will match generic strings like `.a/` or `.x/` instead of a literal `../`.

srafi1:
Adds support for path autocompletion using environment variables such as `$HOME` and `$PWD`. Can be used in the following manner: `$PWD/tests/`.

Fixes #320, fixes #322.

### DIFF
--- a/lua/compe_path/init.lua
+++ b/lua/compe_path/init.lua
@@ -66,7 +66,7 @@ Source._dirname = function(self, context)
   local prefix = string.sub(context.before_line, 1, s + 1)
 
   local buf_dirname = vim.fn.expand(('#%d:p:h'):format(context.bufnr))
-  if prefix:match('%../$') then
+  if prefix:match('%.%./$') then
     return vim.fn.resolve(buf_dirname .. '/../' .. dirname)
   elseif prefix:match('%./$') then
     return vim.fn.resolve(buf_dirname .. '/' .. dirname)
--- a/lua/compe_path/init.lua
+++ b/lua/compe_path/init.lua
@@ -72,6 +72,8 @@ Source._dirname = function(self, context)
     return vim.fn.resolve(buf_dirname .. '/' .. dirname)
   elseif prefix:match('~/$') then
     return vim.fn.expand('~/' .. dirname)
+  elseif prefix:match('%$[%a_]+/$') then
+    return vim.fn.expand(prefix:match('%$[%a_]+/$') .. dirname)
   elseif prefix:match('/$') then
     local accept = true
     -- Ignore URL components
--- a/lua/compe_path/init.lua
+++ b/lua/compe_path/init.lua
@@ -2,7 +2,7 @@ local compe = require'compe'
 
 -- TODO: ' or " or ` is valid as filename
 local NAME_PATTERN = [[\%([^/\\:\*?<>'"`\|]\)]]
-local DIRNAME_REGEX = vim.regex(([[\%(/PAT\+\)*\ze/PAT*$]]):gsub('PAT', NAME_PATTERN))
+local DIRNAME_REGEX = vim.regex(([[\%(\$PAT\+\)\?\%(/PAT\+\)*\ze/PAT*$]]):gsub('PAT', NAME_PATTERN))
 local MENU = {
   DIR = '[Dir]',
   FILE = '[File]',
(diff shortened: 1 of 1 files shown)


lua/compe_path/init.lua
```diff
@@ -2,7 +2,7 @@ local compe = require'compe'
 
 -- TODO: ' or " or ` is valid as filename
 local NAME_PATTERN = [[\%([^/\\:\*?<>'"`\|]\)]]
-local DIRNAME_REGEX = vim.regex(([[\%(/PAT\+\)*\ze/PAT*$]]):gsub('PAT', NAME_PATTERN))
+local DIRNAME_REGEX = vim.regex(([[\%(\$PAT\+\)\?\%(/PAT\+\)*\ze/PAT*$]]):gsub('PAT', NAME_PATTERN))
```
hrsh7th, line 5: We shouldn't modify it.

Sorry for the confusing, but the current implementation is handle `$HOME/.config/nvim/init.vim` as `{ prefix = '$HOME/', dirname = '.config/nvim/init.vim' }`

So we don't have to include `$HOME` as dirname.

lua/compe_path/init.lua
```diff
@@ -66,12 +66,14 @@ Source._dirname = function(self, context)
   local prefix = string.sub(context.before_line, 1, s + 1)
 
   local buf_dirname = vim.fn.expand(('#%d:p:h'):format(context.bufnr))
-  if prefix:match('%../$') then
+  if prefix:match('%.%./$') then
```
hrsh7th, line 69: Thanks!

lua/compe_path/init.lua
```diff
@@ -66,12 +66,14 @@ Source._dirname = function(self, context)
...
     return vim.fn.resolve(buf_dirname .. '/../' .. dirname)
   elseif prefix:match('%./$') then
     return vim.fn.resolve(buf_dirname .. '/' .. dirname)
   elseif prefix:match('~/$') then
     return vim.fn.expand('~/' .. dirname)
+  elseif prefix:match('^%$$') then
+    return vim.fn.expand(dirname)
```
hrsh7th, line 76:
```
  elseif prefix:match('%$[%a_]+/$') then
    return vim.fn.expand(prefix:match('%$[%a_]+/$') .. dirname)
```

lua/compe_path/init.lua
srafi1: I see, I added what you sent in your other comment.

srafi1:
Note that typing something like `$NONEXISTENT_VAR/` will evaluate to `/` because the environment variable does not exist, so it will autocomplete from the `/`. As far as I am concerned, that's fine, but let me know if you want to change that.

hrsh7th:
> Note that typing something like $NONEXISTENT_VAR/ will evaluate to / because the environment variable does not exist, so it will autocomplete from the /. As far as I am concerned, that's fine, but let me know if you want to change that.

Thank you for your pointed out. I hope it can be fixed.

hrsh7th:
But it looks good to me so I can merge it.

hrsh7th:
Thank you very much for your contribution.

I've merge it for now. Probably, we can improve the source but it can be another separated PR.

PR #337, opened 2021-04-16
Add UltiSnips documentation preview

### LINKED ISSUE #327: ultisnips interpolation
I'm using UltiSnips as one of the sources:
```
require'compe'.setup {                           
  preselect = 'disable',                         
  source_timeout = 100,                          
                                                 
  source = {                                     
    path = {kind = ""},                         
    buffer = {kind = ""},                       
    calc = {kind = ""},                         
    nvim_lsp = {kind = ""},                     
    nvim_lua = {kind = ""},                     
    spell = {kind = ""},                        
    ultisnips = {kind = ""},                    
    tabnine = {kind = "", priority = 50},       
    emoji = {kind = "ﲃ", filetypes={"markdown"}},
  },                                             
}                                                
```

And `compe` doesn't show what is inside the snippet.
image
It is also doesn't expand the snippet: if I press `<CR>` it just paste `addaarg`.

Is it possible to achieve `CoC` way of doing this?
image

baldore:
## Description
Closes #327. Adds logic to show UltiSnips preview when selecting a snippet.

## Demo
demo

### DIFF
--- a/lua/compe_ultisnips/init.lua
+++ b/lua/compe_ultisnips/init.lua
@@ -2,6 +2,32 @@ local compe = require'compe'
 
 local M = {}
 
+local function get_snippet_preview(data, args)
+  local filepath = string.gsub(data.location, '.snippets:%d*', '.snippets')
+  local _, _, linenr = string.find(data.location, ':(%d+)')
+  local content = vim.fn.readfile(filepath)
+
+  local snippet = {}
+  local count = 0
+
+  table.insert(snippet, '```' .. args.context.filetype)
+  for i, line in pairs(content) do
+    if i > linenr - 1 then
+      local is_snippet_header = line:find('^snippet%s[^%s]') ~= nil
+      count = count + 1
+      if line:find('^endsnippet') ~= nil or is_snippet_header and count ~= 1 then
+        break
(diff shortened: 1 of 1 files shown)


lua/compe_ultisnips/init.lua
```diff
@@ -15,13 +41,17 @@ function M:determine(context)
 end
 
 function M:complete(args)
-  local received_snippets = vim.F.npcall(vim.call, 'UltiSnips#SnippetsInCurrentScope') or {}
+  local received_snippets = vim.F.npcall(vim.call, 'UltiSnips#SnippetsInCurrentScope', 1) or {}
```
hrsh7th, line 44: We should use `vim.g.current_ulti_dict_info` instead of it?

baldore: I now remember why I didn't use it. I tried it when checking that the table was empty and it returned false, because it had some weird value. Something like [true]: 6.

That's why I left the first variable to check that it was empty, and then, using the global variable with the additional info.

hrsh7th: OK. I've understand.

PR #364, opened 2021-05-06
add emojis

McAuleyPenney:


hrsh7th:
The emojis was created by Lua script https://github.com/hrsh7th/nvim-compe/blob/master/lua/compe_emoji/update.lua.

So I don't prefer the manual adding. Sorry.

Issue #374, opened 2021-05-13
slow on startup


siduck76:
nvim-compe takes so much time on startup time. ( I know it depends on hardware too but it takes more time than coc or completion.nvim  )

nvim-compe on my old pc , with hdd 
![image](https://user-images.githubusercontent.com/59060246/118135779-0a05ba00-b421-11eb-87bf-55a3a25c1cff.png)

main pc on ssd 
![image](https://user-images.githubusercontent.com/59060246/118136073-581abd80-b421-11eb-842a-f8209798476f.png)

my config : https://github.com/siduck76/neovim-dotfiles/blob/main/lua/compe-completion.lua

ii14:
Slow startup is related to #369

For now what you can do is manually disabling sources you're not using.
For example, if you're using buffer, calc, nvim_lsp, path and ultisnips:
```vim
" let g:loaded_compe_buffer = 1
" let g:loaded_compe_calc = 1
let g:loaded_compe_emoji = 1
let g:loaded_compe_luasnip = 1
" let g:loaded_compe_nvim_lsp = 1
let g:loaded_compe_nvim_lua = 1
let g:loaded_compe_omni = 1
" let g:loaded_compe_path = 1
let g:loaded_compe_snippets_nvim = 1
let g:loaded_compe_spell = 1
let g:loaded_compe_tags = 1
let g:loaded_compe_treesitter = 1
" let g:loaded_compe_ultisnips = 1
let g:loaded_compe_vim_lsc = 1
let g:loaded_compe_vim_lsp = 1
let g:loaded_compe_vsnip = 1
```

siduck76:
@ii14 I use nvim-lsp tho so compi_omni shouldnt be disabled too right?

ii14:
You don't need omni for neovim's LSP. I think it's just for plugins that support omnicompletion only

siduck76:
thanks a lot buddy , this really helps . I just use neovim lsp and vsnip+friendly-snippets . This made my neovim even more faster :D 

![image](https://user-images.githubusercontent.com/59060246/118344976-26077980-b54f-11eb-9e4e-3db2720b9a59.png)

hrsh7th:
https://github.com/hrsh7th/nvim-compe/pull/376

siduck76:
> Slow startup is related to #369
> 
> For now what you can do is manually disabling sources you're not using.
> For example, if you're using buffer, calc, nvim_lsp, path and ultisnips:
> 
[...]

Issue #369, opened 2021-05-11
Make sources from /after/plugin/ lazyloading


monkoose:
#### Is your feature request related to a problem? Please describe.
Currently even so I don't use half of sources like `omni, treesitter` etc they are still loading on nvim startup. Could it be improved?
You can disable them with explicitly specify like `let g:loaded_compe_omni = 0`, but isn't it too verbose?

#### Describe the solution you'd like
Load them only on demand when you add them in your `source` option.

ii14:
It's a duplicate of #220, which got closed by the bot. @hrsh7th suggested separating the sources into separate repos, but I think that just as you say, lazy loading is a better alternative and should be a pretty straightforward thing to do. I'll try to do it over the weekend, because it does improve startup time by a lot.

PR #376, opened 2021-05-15
Lazy loading

### LINKED ISSUE #220: Improve start-up time?
I'm experimenting with nvim-compe, and notice that it makes a modest dent in my start-up time.  (Not much, but, still!)

I notice that we are sourcing and registering all of these sources, although I am not using most of them:
```
116.734  000.989  000.989: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_buffer.vim
117.543  000.793  000.793: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_calc.vim
118.461  000.901  000.901: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_nvim_lsp.vim
119.273  000.794  000.794: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_nvim_lua.vim
119.995  000.707  000.707: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_omni.vim
120.839  000.829  000.829: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_path.vim
121.607  000.751  000.751: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_snippets_nvim.vim
122.311  000.686  000.686: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_spell.vim
123.017  000.690  000.690: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_tags.vim
124.047  001.015  001.015: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_treesitter.vim
124.948  000.884  000.884: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_ultisnips.vim
125.008  000.044  000.044: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_vim_lsc.vim
125.038  000.013  000.013: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_vim_lsp.vim
125.794  000.744  000.744: sourcing /home/dch/.vim/plugged/nvim-compe/after/plugin/compe_vsnip.vim
```
Perhaps it would be worthwhile to arrange that these things are loaded only if the configuration says that they will be used?

ii14:
Implements basic lazy loading.

If a source is requested before `after/plugin` files are evaluated, so for example by `g:compe` dictionary, the source is registered in a table that defers loading to `after/plugin/compe.vim`. After that requested sources are going to be registered right when the user calls `compe#setup()`.

In my case it cut down compe loading time in half.

I left the old `after/plugin` files in `after_` directory for now.

Fixes #220 #369 #374


after/plugin/compe.vim
```diff
@@ -0,0 +1 @@
+lua require'compe.lazy'.load_deferred()
```
ii14, line 1: `Lazy.load_deferred` can only be called once, so the `exist(g:compe_after_loaded)` guard shouldn't be necessary. Unless someone wants to prevent that function being called, but I can't think of a single reason why anyone would want that.

lua/compe/lazy.lua
```diff
@@ -0,0 +1,109 @@
+local Lazy = {}
+
+Lazy._sources = {
```
ii14, line 3: Contains only builtin sources. If someone installs a 3rd party source then the chances are he will actually want that source to be loaded anyways, so we don't have to change anything else, like exposing it in public API and letting people add their own stuff here. You just won't pay for the builtin sources you don't use.

lua/compe/lazy.lua
```diff
@@ -0,0 +1,109 @@
...
+  nvim_lsp = function()
+    require'compe_nvim_lsp'.attach()
+  end,
+
+  nvim_lua = function()
+    require'compe'.register_source('nvim_lua', require'compe_nvim_lua')
+  end,
+
+  ultisnips = function()
+    if vim.g.did_plugin_ultisnips ~= nil then
```
ii14, line 64: These functions are getting nulled after they've been called, but maybe there should be a `g:loaded_compe_ultisnips` guard instead, just like it used to work in vim script?

ii14:
@hrsh7th, what do you think about this solution?

hrsh7th:
Basicaly, I don't prefer this solution but I think it is good for current compe.

hrsh7th:
In theory, we should separate all sources (buffer, path, calc, ultisnips, vsnip etc etc...).

hrsh7th:
LGTM.

ii14:
Thanks.

If we want to separate them, we could use autoload like some other plugins do. For example, let's say you want to add a `buffer` source. We could dynamically build a function name like `compe#source#buffer#register()` and vim will try to load it from `autoload/compe/source/buffer.vim`.

Oops, #369 and #374 weren't automatically closed. Can you close them?

hrsh7th:
OK. I will close it.

I think the lazy load is not need if we separate all sources (because the user will install only the using sources.

PR #378, opened 2021-05-16
Add custom formatting functions

elkowar:
This PR adds new configuration options to allow advanced users to configure everything in how their results and documentation get's displayed. It does this by adding a new field to the configuration table.
This may, for example, be used to achieve the feature requested in #209, by adding a simple function (which is used as an example in the readme).
That configuration example results in the list as shown on this screenshot:
image


elkowar:
The provided example config is still rather suboptimal right now, I'll fix that later.

elkowar:
Done and ready for review!

kamalmarhubi:
This is not "an issue", it's a PR that someone spent time on. Opened #526 to change the bot's configuration.

PR #390, opened 2021-05-29
Always filter using filter_text when provided

### LINKED ISSUE #375: Matching completion items in the middle or suffix

I'm a maintainer of an LSP server for a [Markdown-based notebook](https://github.com/mickael-menu/zk) and trying to figure out the best possible config for Neovim's built-in LSP client.

I'm using auto-completion to create links to other notes. Unfortunately, with nvim-compe I have to type the first letter (prefix) of the completion item otherwise it won't match. Here's a screencast showing the problem.

You can see that I can't match "A future-proof..." with `fp`, I need to type `afp`.

https://user-images.githubusercontent.com/58686775/118142862-1121cd80-b40b-11eb-94be-41a25bcc43b2.mov

It would be helpful to match in the middle or end of completion items too, possibly with a lower score than prefix matching. This is the behavior I have using coc.nvim, for example.

This feature could be exposed as an additional option to not change the current behavior.

Apart from that I really enjoyed using nvim-compe which has the most efficient completion I tried, thanks for the effort you put in!

mickael-menu:
Fix #375 

If the filtering user input didn't start with the first character of the `filter_text`, the filter text was ignored.

### DIFF
--- a/lua/compe/matcher.lua
+++ b/lua/compe/matcher.lua
@@ -15,9 +15,7 @@ Matcher.match = function(context, source, items)
     local word = item.original_word
     if #input > 0 then
       if item.filter_text and #item.filter_text > 0 then
-        if Character.match(string.byte(input, 1), string.byte(item.filter_text, 1)) then
-          word = item.filter_text
-        end
+        word = item.filter_text
       end
     end
 


lua/compe/matcher.lua
```diff
@@ -15,9 +15,7 @@ Matcher.match = function(context, source, items)
     local word = item.original_word
     if #input > 0 then
       if item.filter_text and #item.filter_text > 0 then
-        if Character.match(string.byte(input, 1), string.byte(item.filter_text, 1)) then
-          word = item.filter_text
-        end
+        word = item.filter_text
       end
```
hrsh7th, line 19: I think we should respect `prefix matched abbr` as first priority.
So I'd create another PR (#392). Please check it.

mickael-menu:
Superseded by https://github.com/hrsh7th/nvim-compe/pull/392

PR #392, opened 2021-05-30
Choose word more smarter

### LINKED ISSUE #375: Matching completion items in the middle or suffix

hrsh7th:
Fix #375 

Related #390

### DIFF
--- a/lua/compe/matcher.lua
+++ b/lua/compe/matcher.lua
@@ -1,4 +1,5 @@
 local Character = require'compe.utils.character'
+local String = require'compe.utils.string'
 
 local Matcher = {}
 
@@ -13,17 +14,13 @@ Matcher.match = function(context, source, items)
     item.index = i
 
     local word = nil
-    if #input > 0 then
-      for _, key in ipairs({ 'original_abbr', 'filter_text', 'original_word' }) do
-        if item[key] and #item[key] > 0 then
-          word = item[key]
-          if Character.match(string.byte(input, 1), string.byte(item[key], 1)) then
-            break
-          end
+    for _, key in ipairs({ 'original_word', 'filter_text', 'original_abbr' }) do
+      if item[key] and #item[key] > 0 then
+        word = item[key]
+        if String.match_prefix(item[key], input) then
+          break
         end
       end
-    else
-      word = item.original_word
     end
 
     if #word >= #input then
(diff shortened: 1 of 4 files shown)


hrsh7th:
@mickael-menu What do you think about this PR?

lua/compe/matcher.lua
```diff
@@ -12,13 +12,18 @@ Matcher.match = function(context, source, items)
   for i, item in ipairs(items) do
     item.index = i
 
-    local word = item.original_word
+    local word = nil
     if #input > 0 then
-      if item.filter_text and #item.filter_text > 0 then
-        if Character.match(string.byte(input, 1), string.byte(item.filter_text, 1)) then
-          word = item.filter_text
+      for _, key in ipairs({ 'original_abbr', 'filter_text', 'original_word' }) do
```
mickael-menu, line 17: @hrsh7th It doesn't fix #375 for me , because in my case none of `original_abbr`, `filter_text` or `original_word` are matching the prefix, so it falls back on the last one in the list: `original_word`.

If I set `filter_text` as the last item though, it fixes the issue:


```suggestion
      for _, key in ipairs({ 'original_abbr', 'original_word', 'filter_text' }) do
```

hrsh7th: Ah! Thanks. It's my bad.

I will consider it more deeply...

(I want to prefer `prefix matched abbr` as first priority)

mickael-menu: That's fine, if you do this fix it works for me, but if you want to do it differently, let me know so I can test again.

danymat:
@hrsh7th Hello, it works for me !

mickael-menu:
It's working for me too, thanks 👍

PR #409, opened 2021-06-10
faq - document ESC + vim-autoclose. Fixes #408

### LINKED ISSUE #408: Document how to close with the ESC key

My problem is that I was expecting that pressing ESC would close the autocomplete window (I've seen that referred as ... "pug" in the issues?). It does not, by default. But the suggested mappings in the doc don't explain how to accomplish this.

I am not super familiar with the different ways Vim allows mapping keys (imap vs innoremap) so perhaps I am missing something on that side.

I'd like the docs to include a mapping that makes ESC close the autocomplete window.

#### Describe alternatives you've considered

I tried doing this:

```
inoremap <silent><expr> <ESC>     compe#close('<ESC>')
```

It did not go well because:
* It did not close the window
* It showed an error message 
* It made it impossible to go back to normal mode in Vim xD. I had to edit my init.vim file in nano in order to rescue it.

#### Additional context

As an extra, I usually don't use `<ESC>`, but a combination of `j` and `k`. So I have this on my conf:

```
imap jk <ESC>
imap Jk <ESC>
imap JK <ESC>
```

Ideally (for me) the provided mapping should be compatible with that (so pressing `jk` in rapid succession would close the autocomplete window, but otherwise individual presses of `j` and `k` allow selecting from alternatives on it.

kikito:


hrsh7th:
Thank you!

PR #414, opened 2021-06-15
Evaluate Interpolated Expressions (UltiSnips)

### LINKED ISSUE #384: Feature: UltiSnips preview should evaluate expressions

It would be nice if the documentation preview for UltiSnips implemented in #337 evaluated expressions as denoted by backticks (`) instead of showing the expression.

Given this snippet:
```
snippet Example
    Date: `strftime("%Y-%m-%d")`
endsnippet
```

Currently, the documentation preview would show the following text:
```
Date: `strftime("%Y-%m-%d")`
```

The requested feature would show the following instead by evaluating `strftime()`:
```
Date: 2021-05-20
```

#### Describe alternatives you've considered

coc-snippets does this (and that's where I got the idea from).

#### Additional context

UltiSnips supports both python and vimscript by putting the following in backticks, respectively: `!p # python code` or `!v " vimscript code`. However, those are beyond the scope of this request.

I can try to implement this when I have time next week, but first I want to make sure that it is a feature more people would like to have.

srafi1:
UltiSnips supports interpolated expressions as documented [here](https://github.com/SirVer/ultisnips/blob/master/doc/UltiSnips.txt#L820-L1069). This PR evaluates these expressions for the documentation preview of the snippet. Shell code and Vimscript expressions are implemented, but not Python code as it is much more involved and I'm not sure how to approach it.

Example:
With the snippet:
```
snippet date
`date "+%d-%m-%Y"`
endsnippet
```

Current behavior:
DeepinScreenshot_select-area_20210614222456

New behavior:
DeepinScreenshot_select-area_20210614222407

Fixes #384.

### DIFF
--- a/lua/compe_ultisnips/init.lua
+++ b/lua/compe_ultisnips/init.lua
@@ -25,7 +25,16 @@ local function get_snippet_preview(data, args)
         local s, e = EXPRESSION_REGEX:match_str(line)
         while s do
           local expr = string.sub(line, s+2, e-1)
-          local evaluated = vim.api.nvim_eval(expr)
+          local prefix = string.sub(expr, 1, 2)
+          local evaluated
+          if prefix == '!v' then
+            evaluated = vim.api.nvim_eval(string.sub(expr, 3))
+          elseif prefix == '!p' then
+            -- TODO: implement python interpolation (value of snip.rv)
+            evaluated = expr
+          else
+            evaluated = vim.fn.system(expr)
+          end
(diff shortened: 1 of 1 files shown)


hrsh7th:
Hm... I can't maintain it... if ultisnips does not provide similar feature, we must not realize it on compe side.

or you can take over `compe-ultisnip` to your repo.

srafi1:
Fair enough. I will consider creating my own plugin as a ultisnips source, but, at this point, I think it would be easier for me personally to just switch over to vsnip or any other snippets engine that's better supported by nvim-compe already.

PR #420, opened 2021-06-19
fix default throttle time

Diaoul:
I've noticed that [the docs](https://github.com/hrsh7th/nvim-compe/blob/master/doc/compe.txt#L257-L260) and the [README](https://github.com/hrsh7th/nvim-compe/blob/master/README.md) say that throttle time defaults to 80 but [the code](https://github.com/hrsh7th/nvim-compe/blob/master/lua/compe/config.lua#L4) says otherwise. This is a fix for that.


hrsh7th:
Thank you

PR #433, opened 2021-06-27
Fixes #432

### LINKED ISSUE #432: Documentation Window Positioned incorrectly with multigrid extension
#### Checkhealth
```
health#compe#check
========================================================================
## compe:snippet
  - INFO: snippet engine is not detected.

## compe:mapping
  - INFO: `compe#complete` is not mapped
  - INFO: `compe#confirm` is not mapped
  - INFO: `compe#close` is not mapped
  - INFO: `compe#scroll` is not mapped
```

#### Describe the bug
When using a gui with multigrid enabled, windows are positioned relative to the entire editor rather than the current window leading to overlapping windows.

#### To Reproduce

1. Use a multigrid enabled gui such as fvim or neovide
2. Open two windows with a buffer that can show the documentation window
3. Notice that when the documentation is open on the left window, the positioning is fine because that window has no offset.
4. Notice that when the documentation is open on the right window, it is positioned as though the cursor were actually on the left window because it ignores the position of the window

#### Screenshots (optional)
image

#### Additional context (optional)
It is possible that this is actually an issue with the multigrid implementations of fvim and neovide, but I noticed that the floating window position received by the gui is relative to the editor and not a given window, so I think it has to be an issue with how the window is constructed.

Kethku:
Fixes https://github.com/hrsh7th/nvim-compe/issues/432 by modifying the floating window utility functions to use relative "win" rather than relative "editor". This is because in multigrid environments the position of the documentation window was incorrect as explained in the issue.

### DIFF
--- a/autoload/compe/documentation.vim
+++ b/autoload/compe/documentation.vim
@@ -54,6 +54,7 @@ function! compe#documentation#open(text) abort
   endif
 
   silent call s:window.open({
+  \   'relative': 'win',
   \   'row': l:pos[0] + 1,
   \   'col': l:pos[1] + 1,
   \   'width': l:state.size.width,
--- a/autoload/vital/_compe/VS/Vim/Window/FloatingWindow.vim
+++ b/autoload/vital/_compe/VS/Vim/Window/FloatingWindow.vim
@@ -213,6 +213,10 @@ function! s:FloatingWindow.open(args) abort
   \   'origin': get(a:args, 'origin', 'topleft'),
   \ }
 
+  if has('nvim')
+    l:style.relative = get(a:args, 'relative', 'editor)
+  endif
+
   let l:will_move = self.is_visible()
   if l:will_move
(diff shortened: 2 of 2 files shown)


Kethku:
Actually looks like this breaks non multigrid environments... I'll have to think a bit

Kethku:
I'm not sure what the right solution is. Looks like pum_getpos returns relative positions when multigrid. I could special case whether editor or win is used for the relative property depending on if a gui is connected, but that feels like a hack.

Any guidance would be appreciated

hrsh7th:
The big change that relates to the documentation window had merged. Could you check the problem with latest compe?

Kethku:
Finally got around to testing this. Unfortunately its still broken. In a multigrid environment, the documentation window is positioned relative to the top window not the current window 
image

Kethku:
Changing the relative option in the window creation to win fixes this issue for multigrid environments, but I believe breaks it for normal ones. I'm not sure what the correct way to solves this is for both use cases.

PR #434, opened 2021-06-28
Make spell completion respect vim spell option

adamreese:
This makes spell completion respect `set nospell`.

The current conditional `vim.wo.spell == 0` will always return false.


hrsh7th:
Oh... vim.wo.* returns boolean instead of vim's number. Thank you!

PR #438, opened 2021-07-01
ref(README): simplify `check_back_space`

Iron-E:
For this function, using an `if` statement to `return` a boolean value is the same as just returning that value directly. The one caveat is that `:match` doesn't return a bool value, so we have to convert it into a boolean value by testing for `~= nil`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,11 +310,7 @@ end
 
 local check_back_space = function()
     local col = vim.fn.col('.') - 1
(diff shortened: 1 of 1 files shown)


Iron-E:
Oops. I just noticed I forgot to put the `vim.` before `fn.getline`. I'll amend the previous commit to fix that tomorrow.

hrsh7th:
LGTM. Thank you!

PR #439, opened 2021-07-01
Added more documentation info for tags

00sapo:
Actually, I would like to put `tag.kind` in the `menu` entry near the `[T]` symbol (e.g. `f [T]` as a symbol for a function would be nice. I don't know how to achieve that, though...

Other things that could be added is the context from the file around the specific line, if the line is available; it could substitute the exe cmd used to find the tag

### DIFF
--- a/lua/compe_tags/init.lua
+++ b/lua/compe_tags/init.lua
@@ -40,20 +40,21 @@ function Source.documentation(_, context)
       break
     end
     local doc =  tag.filename .. ' [' .. tag.kind .. ']'
-    if #tag.cmd >= 5 then
-        doc = doc .. '\n  ' .. tag.cmd:sub(3, -3):gsub('%s+', ' ')
+        local doc =  '# ' .. tag.filename .. ' [' .. tag.kind .. ']'
+    if #tag.cmd >= 5 and tag.signature == nil then
+        doc = doc .. '\n  __' .. tag.cmd:sub(3, -3):gsub('%s+', ' ') .. '__'
     end
     if tag.access ~= nil then
         doc = doc .. '\n  ' .. tag.access
     end
     if tag.implementation ~= nil then
-        doc = doc .. '\n  ' .. tag.implementation
+        doc = doc .. '\n  impl: _' .. tag.implementation .. '_'
     end
     if tag.inherits ~= nil then
         doc = doc .. '\n  ' .. tag.inherits
     end
     if tag.signature ~= nil then
-        doc = doc .. '\n  ' .. tag.signature
+        doc = doc .. '\n  sign: _' .. tag.name .. tag.signature .. '_'
     end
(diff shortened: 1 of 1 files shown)


hrsh7th:
Sorry. Actualy, I don't use ctags so I can't debug it.
Could you please provide the animated gif fo this?

00sapo:
Peek 2021-07-01 15-06

hrsh7th:
Looks very good!

You can use markdown syntax for documents.
I think it would be better if the filename line as a markdown header.

```
local doc =  '#' .. tag.filename .. ' [' .. tag.kind .. ']'
```

00sapo:
Thanks, added it! Can I ask what is the argument in `get_metadata`? I would like to put signature information and change the `[T]` if we have information. The signature could then be used by signature help plugins...

00sapo:
Peek 2021-07-01 15-06

hrsh7th:
nvim-compe does not relate to signature help so your suggestion doesn't possible, sorry.

the syntax looks good! 👍🏼

hrsh7th:
I will merge it if you have no additional improvements.

00sapo:
> nvim-compe does not relate to signature help so your suggestion doesn't possible, sorry.
> 
> the syntax looks good! 👍🏼

I will try another way then, and in case I'll write here again. There's no way to change `[T]` based on the metadata extracted from tag information? (the kind of tag?)

hrsh7th:
The `[T]` means `[Tag]`? If so, you can change the below codes.
https://github.com/hrsh7th/nvim-compe/blob/130d88edb551cb97c73e4ba19d8aca458b07439a/lua/compe_tags/init.lua#L21

```
  local _, items = pcall(function()
    local items = {}
    for _, tag in ipairs(vim.fn.getcompletion(context.input, "tag")) do
      tag.menu = '[Modified Tag]'
      table.insert(items)
    end
    return items
  end)
```

00sapo:
Your code doesn't work for me, but I'm not expert in vim plugins. Can you point me to how to debug lua scripts? Can I read some log?

hrsh7th:
Sorry. it's easy mistake.

```
  local _, items = pcall(function()
    local items = {}
    for _, tag in ipairs(vim.fn.getcompletion(context.input, "tag")) do
      tag.menu = '[Modified Tag]'
      table.insert(items, tag)
    end
    return items
  end)```

The above code is fixed version.

00sapo:
Still not working...

hrsh7th:
You should remove `pcall` and set `let g:compe.debug = v:true` (these suppress all error messages).

00sapo:
Sorry, I can't get it working. Maybe in future I'll come back on this. I think you can merge, then

hrsh7th:
Thank you. OK. I merge it.

PR #441, opened 2021-07-01
ref(README): call `vsnip#` functions directly

Iron-E:
Previously, there was a `vim.fn.call` invocation which was used to call vsnip autoload functions (assumedly, because they use the special `#` syntax). However, this can be done with `vim.fn['vsnip#foo']` directly.


hrsh7th:
LGTM~.

PR #446, opened 2021-07-05
fix(documentation): check if doc is table before deep copy

folke:
Fixup for my previous PR.

This adds a check if the doc is a table. Only then do we make a deep copy. Would fail otherwise for some lsp servers


hrsh7th:
Sorry. I was missed it. Thank you.

PR #455, opened 2021-07-08
fix: change LuaSnip's menu field to LuaSnip

### LINKED ISSUE #454: Questions regarding LuaSnip source
Hey, I have noticed that LuaSnip has pretty different representation on completion menu.

Screenshot 2021-07-08 at 00 00 09

1) I have noticed that vsnip has menu set to`[Vsnip]`, snippets.nvim sets `[Snippets]` (presumably due to its name), Ultisnips sets `[Ultisnips]` but luasnip has very different `[Snippets]`. I know it's very easy to change it in user's configuration but perhaps it should be kept analogical between sources. Would you accept PR changing it to `[LuaSnip]`?

2) `kind` is set to filetype (or `all` perhaps, but I don't have those snippets enabled so am not sure), while vsnip does not set it at all, bot ultisnip and snippets.nvim set it to `Snippet`. What was the reasoning behind having filetype in there?

3) And finally, I would really like to render luasnip's snippets in documentation window. Could that perhaps be configurable if it's not widely desired?  

CC @L3MON4D3

gegoune:
Partialy fixes #454

### DIFF
--- a/lua/compe_luasnip/init.lua
+++ b/lua/compe_luasnip/init.lua
@@ -12,7 +12,7 @@ function Source.get_metadata(_)
     priority = 10,
     -- keep Snippets with same trigger but different filetype.
     dup = true,
-    menu = "[Snippets]"
+    menu = "[LuaSnip]"
   }
 end
 


hrsh7th:
LGTM. Thank you!

PR #457, opened 2021-07-08
Account for multiline descriptions in luasnip-source.

L3MON4D3:
LSP-snippets may have tables as description, which causes an error here.


hrsh7th:
Thank you.

PR #477, opened 2021-07-14
fix: don't show empty doc for lsp

folke:
Don't show empty docs.

This actually led to an error in the float module since we tried rendering a float of width `0`


hrsh7th:
Thsnk you!

PR #478, opened 2021-07-14
DateTime module

princebett:
Hello. This is a simple module to autocomplete the date and time. Source.confirm feels hacky so if there's any way it can 
be improved I'd like some help. I'm new to writing nvim lua code. Awesome plugin !

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,7 @@ highlight link CompeDocumentation NormalFloat
 
 - buffer
 - path
+- datetime
 - tags
 - spell
 - calc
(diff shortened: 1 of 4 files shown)


gegoune:
It looks to me like that belongs more in snippets engine rather than completion one.

princebett:
How different is it from Spell ? Is there a snippet engine with the functionality built in ?

hrsh7th:
Yes. Current, the implementatuon should be realized by snippet engine.

PR #516, opened 2021-08-09
Fixes #400

### LINKED ISSUE #400: Ultisnips 'kind' setting
#### Describe the bug

Using Ultisnips for snippets does not allow the user to change 'kind'.

#### To Reproduce

Setting the 'menu' value for the Ultisnips source works fine (as do other sources) but (I believe) Compe overwrites the 'kind' value for Ultisnips.

In the example below i change the 'menu'-item to '[US]' which works great and then the 'kind'-value to anything.

```lua
compe.setup {
  enabled = true;
  autocomplete = true;
  debug = false;
  min_length = 1;
  preselect = 'enable';
  throttle_time = 80;
  source_timeout = 200;
  incomplete_delay = 400;
  max_abbr_width = 100;
  max_kind_width = 100;
  max_menu_width = 100;
  documentation = true;

  source = {
    path = true;
    buffer = {menu='[B]'};
    calc = true;
    nvim_lsp = true;
    nvim_lua = true;
    ultisnips = {menu='[US]', kind='WHATEVER'}; 
}
```

#### Actual behavior

The 'kind' value for Ultisnips is always set to 'Snippet'.

#### Expected behavior

The user defined value of 'kind' is used.

#### Screenshots (optional)

compe-us

#### Additional context (optional)

I believe the problem could be in the [Ultisnips-settings](https://github.com/hrsh7th/nvim-compe/blob/888d9eccfe7b9092dcbfcdfabfad601ce63f3eb5/lua/compe_ultisnips/init.lua) on line 29. Here compe seems to always define the kind of Ultisnips as 'Snippet'.

Cheerio!

StenSipma:
Fixed by passing the metadata as one of the arguments to the 'source.complete' function. Then, check if 'kind' is set in the metadata. The default is still 'Snippet'.

### DIFF
--- a/lua/compe/source.lua
+++ b/lua/compe/source.lua
@@ -149,6 +149,7 @@ Source.trigger = function(self, context, callback)
 
     -- Completion
     self.source:complete({
+      metadata = self.metadata;
       context = self.context;
       input = self.context:get_input(state.keyword_pattern_offset);
       keyword_pattern_offset = state.keyword_pattern_offset;
--- a/lua/compe_ultisnips/init.lua
+++ b/lua/compe_ultisnips/init.lua
@@ -51,12 +51,16 @@ function M:complete(args)
   local snippets_list = vim.g.current_ulti_dict_info
 
   local completion_list = {}
+  local kind = 'Snippet'
+  if args.metadata.kind ~= nil then
+     kind = args.metadata.kind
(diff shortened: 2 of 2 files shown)


hrsh7th:
Thank you.,I've understand.

PR #520, opened 2021-08-10
Fix formatting in known-issues section

astier:
All links are displayed in one line. I think it would be better for readability if each link would be displayed on a separate line.


hrsh7th:
Thank you!

PR #538, opened 2021-08-20
allow hidden buffers in completion with option

joshzcold:
This allows completion of words from buffers that aren't currently open if allowed in the configuration.

TODO:
- [ ] ill need some help in determining the best way to insert this configuration option. There didn't seem like a good method for setting options at the source level, so I had to create a global option.
- [ ] update documentation


joshzcold:
Closing in favor of functionality in https://github.com/hrsh7th/cmp-buffer
```lua
          {
              name = 'buffer', 
              opts = {
                  get_bufnrs = function()
                      return vim.api.nvim_list_bufs()
                  end
              }
          },
```
works great 👍

PR #545, opened 2021-10-02
Fix broken Luasnip integration

kdarkhan:
In this change https://github.com/L3MON4D3/LuaSnip/commit/a1580b62beaffad0d631219d1b14764b17ab63aa
LuaSnip removed global table `Luasnip_current_nodes`
which now needs to be accessed as `luasnip.session.current_nodes`


hrsh7th:
thank you.

